PR #3, opened 2020-06-16
Refactor backend itemSearch

nicolengsy:
**Description**
This PR refactors the code in the `items.js`  controller's  `itemSearch()` so that it makes one SQL query to get the set of filtered items to return, rather than getting all item objects with SQL then filtering through them for each filter parameter. Lucas had mentioned I could try doing this when I raised a question about it two weeks ago.

I left the `itemInstanceTag` and `itemTypeTag` out from the SQL query since I'd have to use SQL to search a comma-separated list; can look into it if this refactor is really what we want.

**Improvement**
The performance does not improve much for simple item searches with few parameters (e.g. `api/items/search?status=VERIFIED`) but the time for the request decreases by approx. half for more complex searches (e.g. `/items/search?status=VERIFIED&minPrice=7&maxPrice=12&fairness=true&excludeBeneficiary={username1},{username2}`), which is what happens when a user uses the price filter button (i.e. improves from 1300-1800ms to 600-800ms on my computer).

However this makes the code a little less clean compared to when we keep the SQL code in the `itemHelpers.js` file instead, and the improvement is not significant or noticeable, as I'm not sure yet what the major factor is that makes some pages load slowly (ie. after clicking Help a Family or the price filter button)  Though I notice the `beneficiary/match` search also takes a long time. So I'm not sure if this will make a big difference; this is just a little suggestion since I have the time to try it out! :)


packages/backend/src/server/util/itemHelpers.js
```diff
@@ -153,6 +153,21 @@ async function getVerifiedItemObjsByPriceRange(minPrice, maxPrice) {
   }
 }
 
+async function getItemObjsBySql(condition, params) {
```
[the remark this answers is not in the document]
lucashu1: Actually, this comment is kinda overridden by my higher level comment about creating a general-purpose helper function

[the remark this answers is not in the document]
Raghav-Maheshwari: @lucashu1 @nicolengsy I like the "general-purpose" getItems helper function idea!

nicolengsy:
@lucashu1 The `getItemObjs()` function is complete! The function should be able to cover all the other getter methods in  `itemHelper.js` . I removed some getter methods that I personally created for `itemSearch`. But should I change all instances of the other getter methods to use `getItemObjs()` instead, or simply call `getItemObjs()` within the original getter methods, or just leave them as they are?

For example, calling `getDonatableItems()` seems a lot more intuitive than `getItemObjs({status: 'VERIFIED', beneficiaryIsVisible: 1}) - so I don't know if refactoring those functions away is the most helpful.

lucashu1:
@nicolengsy As a middle ground, let's rewrite `getDonatableItems()` (and other similar getter functions) to call the new all-purpose `getItemObjs()` function. Lmk when that's done, and I'll take another look through the code!

nicolengsy:
@lucashu1 done as requested! This PR also adds a price conversion from USD -> EUR in `itemSearch`, since the previous version seemed to have forgotten to do that (and got away with it because the conversion rate doesn't make a big difference).

packages/backend/src/server/controller/items.js
```diff
@@ -80,34 +81,33 @@ async function itemSearch(req, res) {
...
+
+    // get items
+    const params = (({ status, excludeBeneficiary}) => ({
+      status,
+      minPriceEur: minPriceEur,
+      maxPriceEur: maxPriceEur,
+      excludeBeneficiary,
+      cause: 'refugee',
+      beneficiaryIsVisible: 1,
+    }))(req.query);
```
lucashu1, line 108: I honestly didn't know you could do that with arrow functions, so TIL: https://stackoverflow.com/questions/22138550/immediate-function-using-javascript-es6-arrow-functions

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -66,102 +65,94 @@ function sqlRowToItemObj(row) {
...
-  const [minPriceEur, maxPriceEur] = await Promise.all([
-    currencyHelpers.convertCurrency(minPrice, "USD", "EUR"),
-    currencyHelpers.convertCurrency(maxPrice, "USD", "EUR")
-  ]);
-
+async function getItemObjs({ cause, status, beneficiaryIsVisible, minPriceEur, maxPriceEur, storeId, excludeBeneficiary, itemId, itemIds }={}) {
+  let sqlConditions = '1=1 ';
+  const params = [];
+  // get item objs by id
+  if (typeof itemId !== 'undefined') {
```
lucashu1, line 96: Just to be safe, in each of these `if` statements, let's actually check if it's not `null` as well. So: `typeof itemId !== 'undefined' && itemId !== null`.

But other than that, I think the rest of the PR looks good

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -66,102 +65,94 @@ function sqlRowToItemObj(row) {
...
-    );
-    if (results.length === 0) {
-      return null;
-    }
-    return sqlRowToItemObj(results[0]);
-  } catch (err) {
-    errorHandler.handleError(err, "itemHelpers/getItemObjFromItemId");
-    throw err;
-  }
+  const item = await getItemObjs({ itemId: itemId });
```
lucashu1, line 68: This returns a list, so do `const items = await ...`, and `return items[0]`

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -66,102 +65,94 @@ function sqlRowToItemObj(row) {
...
-    currencyHelpers.convertCurrency(minPrice, "USD", "EUR"),
-    currencyHelpers.convertCurrency(maxPrice, "USD", "EUR")
-  ]);
-
+async function getItemObjs({ cause, status, beneficiaryIsVisible, minPriceEur, maxPriceEur, storeId, excludeBeneficiary, itemId, itemIds }={}) {
+  let sqlConditions = '1=1 ';
+  const params = [];
+  // get item objs by id
+  if (typeof itemId !== 'undefined') {
+    console.log(`itemHelpers.getItemObjs: filtering by itemId = ${itemId}`);
```
lucashu1, line 97: @Raghav-Maheshwari what do you think about keeping/deleting these filtering-related console.logs()? It might clutter up the logs in prod, but I could potentially see them being useful as well

packages/backend/src/server/controller/items.js
nicolengsy: Same! stack overflow saves the day

packages/backend/src/server/util/itemHelpers.js
nicolengsy: Do we want to also do `if (isNan(maxPriceEur)) {...}` checking for min/max price, storeId etc., or still allow for string inputs for those?

lucashu1: `isNaN` actually does allow for string inputs: https://www.w3schools.com/jsref/jsref_isnan.asp

nicolengsy: Oh! Then I'll just change all the number things to `isNan()`!

packages/backend/src/server/util/itemHelpers.js
Raghav-Maheshwari: I'm fine with keeping these as console logs!

PR #10, opened 2020-06-19
Added add-to-cart tooltip

nicolengsy:
This PR adds a tooltip for when a user adds something to cart on a family page, which disappears after 5 seconds.

**Implementation**
An event listener defined at top level `/give/[beneficiaryIdOrUsername]` is passed into a `FamilyProfile`, triggered when an item is added to cart. Chosen item is passed as a prop into the `NavBar` and then into the new component, `CartToolTip`. 

Note: A new button style named `light` was added for the Checkout button. The button style is identical to the button style defined in `ItemCard`, which is why I thought it would be good to make it shared, so that we can later refactor the `ItemCard` component to use this style.


lucashu1:
Launched a review app here! https://duet-mono-website-pr-10.herokuapp.com

Imo, we can make the modal stay a little bit longer? Maybe 10 seconds instead of 5? What do you think @spenceblood @Raghav-Maheshwari

lucashu1:
Also, when I launch the review app in a new window and add an item to cart for the first time, I still get the old modal showing up as well. So we should disable the old modal @nicolengsy 

Screen Shot 2020-06-18 at 10 03 21 PM

lucashu1:
Another bug to squash:
1. Add item to cart
2. Click "remove" from modal
3. Item card still says "Added"! (Should revert to "Add to Donation")

Screen Shot 2020-06-18 at 10 09 55 PM

nicolengsy:
@lucashu1 Oop thanks for catching them! Will make edits tomorrow. From playing around with it I actually think 10 seconds looks rather long, and am wondering if it's best to do a 5 second stay, and to keep it there as long as (or for 5 more seconds) a user hovers over it?

lucashu1:
The hover idea is definitely interesting @nicolengsy! I like that idea too. Maybe play around with it and see what feels most natural? I'll defer to @spenceblood and @Raghav-Maheshwari and your best judgement on this one

spenceblood:
Yes, let's do 5 seconds and have the modal stay on hover. Also, can we add a fade effect so that the opacity goes from 0 → 100 in 200ms? I'm thinking fade in and fade out. It feels a little abrupt right now.  This is looking great Nicole!

packages/frontend/components/shared/NavBar/index.jsx
```diff
@@ -199,16 +204,8 @@ class NavBar extends React.Component {
...
-      const { showShoppingCart } = this.state;
-
       const cartItems = JSON.parse(localStorage.getItem('cartItems'));
       if (cartItems) {
-        if (!showShoppingCart) {
-          this.setState({ showShoppingCartMessage: true });
-          setTimeout(() => {
-            this.setState({ showShoppingCartMessage: false });
-          }, 7000);
-        }
```
Raghav-Maheshwari, line 211: @lucashu1 @nicolengsy curious to hear your guys's thoughts on this.

My feeling is that the new logic for displaying the cart tooltip is a bit complicated and not easily expandable. I.e adding "added item" to the root `familyPage` component state and then passing down the state as a prop into the navbar - if in the future for some reason we want to add items to a cart from a page other than the `familyPage`,  then we have to add a new "added item" state to that page and make sure we're passing it into the navbar.

Could we instead grab the most recently added cart item from the `cart` array inside `localStorage` and render the new tooltip the same way we were rendering the "added to cart" message? I think it would be safe to assume that the last element in the cart array would be the most recently added item (I don't think javascript automatically sorts the cart by item id or something)

We can still have a shared `CartToolTip` component that takes an "added item" as a prop.

Raghav-Maheshwari: Echoing thoughts though that this looks great so far!

lucashu1: Getting the item directly from local storage makes sense to me! I think it'd be fine either way. We'd still need to pass up a boolean variable to indicate whether to show the modal or not, right?

Raghav-Maheshwari: Yeah, we could use a similar `showShoppingCartToolTip` bool to conditionally render the tooltip

nicolengsy:
@Raghav-Maheshwari @lucashu1 the code should be ready for review except for the fade-out effect that I'm figuring out, so you can review it anytime! 

I have 
- refactored the logic as Raghav suggested (no longer passing down a prop but instead using `localStorage` to get the item
- made the modal stay for 5 seconds, then it stays if user hovers over it

For the fade-out effect I'm just trying to see if it can be done completely with CSS or if I need to use javascript - work in progress :)

nicolengsy:
Implemented the fade-out effect. I worked on fading in but commented it out, since it made more sense to me for the modal to pop up when something is added to cart instead of fading in -- would love to know what @spenceblood thinks! I can clean it up after we decide what we want with fading in/out.

The implementation relies on two events in Navbar: `cartUpdate` and `cartRemove`. `cartUpdate` uses the logic @Raghav-Maheshwari suggested where I set a state flag to show the modal; I then fade it out after 5 seconds _only if no new item was added in those 5 seconds, and user is not hovering_. This explains a few more state values i added (`addedItem` and `cartToolTipHover`). 

There is a third new event `updateItemCards` added in `FamilyProfile`, which rerenders the item cards to show the correct button ("Added" or "Add to donation").

I've just been stuck on timing issues since I rely on state values, and `setState()` is asynchronous. Please let me know if there is a better way of implementation!

**Known issues:** There is one last bug, where the modal does not reappear if a user adds an item right after removing it from the modal. I'm working on it but if anyone knows how to fix it I'd love to hear it!

packages/frontend/components/FamilyProfile/index.jsx
```diff
@@ -297,6 +297,13 @@ class FamilyProfile extends React.Component {
         offset: -100,
       });
     }
+
+    // update cart items
+    document.addEventListener('updateItemCards', () => {
+      this.setState({
```
Raghav-Maheshwari, line 303: I don't think there's a need to define a new `updateItemCards` event here. Technically you could just listen in to the `cartUpdate` event and then assume something happened to the cart and therefore update the state of `stateActiveItems`

packages/frontend/components/shared/NavBar/CartToolTip.jsx
```diff
@@ -0,0 +1,167 @@
...
+    const newCartItems = cartItems.filter(itm => itm.itemId !== id);
+    localStorage.setItem('cartItems', JSON.stringify(newCartItems));
+
+    // dispatch event defined in NavBar
+    const cartUpdateEvent = new Event('cartUpdate');
+    document.dispatchEvent(cartUpdateEvent);
+
+    // dispatch event defined in NavBar
+    const cartRemoveEvent = new Event('cartRemove');
+    document.dispatchEvent(cartRemoveEvent);
```
Raghav-Maheshwari, line 99: since all that the `cartRemove` event does is tell the `NavBar` to fire a `updateItemCards` event, either you could do that straight from here, or based on my earlier comment, we could just remove `updateItemCards` and just rely on the `cartUpdate` event. That way the only event we're listening to is the `cartUpdate` event

nicolengsy: Ah, I separated the `cartRemove` event from `cartUpdate` because the checkout page also dispatches the `cartUpdate` event, triggering the tooltip-hiding unnecessarily. I'll put this back inside `cartUpdate` and just add a flag to check if the tooltip needs to be hidden!

packages/frontend/pages/give/[beneficiaryIdOrUsername]/index.jsx
```diff
@@ -186,6 +186,7 @@ class Profile extends React.PureComponent {
             currencyData={currencyData}
             otherFamilies={otherFamilies}
             preselectedItemIds={preselectedItemIds}
+            handleCartToolTip={this.handleCartToolTip}
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/give/[beneficiaryIdOrUsername]/index.jsx:189]: where is handleCartToolTip declared?

nicolengsy:
@Raghav-Maheshwari thank you for the comments! 

Just to restate within this PR, the code is complete besides one bug: when user clicks "Remove" on the tooltip and then "Add to Cart" within 1 second, the tooltip doesn't show up (or disappears too quickly) for the new item. Raghav suggested it has to do with how `setTimeout()` blocks other code execution.

PR #11, opened 2020-06-19
Fix frontend bugs in preparation for monorepo deploy

Raghav-Maheshwari:
This PR fixes some front-end bugs:

- Closed beta subscription now calls `/donations/captureTransaction`
- Fix a UI bug for the twitter share button
- Fix a UI bug on the family page that was caused by the tooltip component


Raghav-Maheshwari:
@lucashu1 I will do some testing on monorepo staging to make sure all of our important donation flows still work! Then it should be ready to be deployed to production

PR #17, opened 2020-06-23
Os typeform integrated

OStouraiti:


packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
```diff
@@ -0,0 +1,65 @@
...
+const FormData = require('form-data');
+
+let fileData = new FormData();
+let { itemId } = {};
+
+export default class FileInput extends React.Component {
+  static async getInitialProps({ query }) {
+    itemId = query;
+    console.log(`!!! ${itemId.itemId}`);
+    return itemId.itemId;
```
lucashu1, line 13: I think it would be cleaner to store `itemId` in the component `prop`s, instead of in a global variable.

So you'd do:
```
static async getInitialProps({ query }) {
  const { itemId } = query;
  return { itemId };
}
```

Then in `handleSubmit`, you can do:
```const { itemId } = this.props;```
in order to access `itemId` from within the `handleSubmit` function.

That way, we can get rid of the `let { itemId }...` line from above. And also, we can simply make `itemId` an int, instead of an object

File: packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+    super(props);
+    this.state = { note: '' };
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.fileInput = React.createRef();
+    this.changeHandler = this.changeHandler.bind(this);
+  }
+
+  componentWillUnmount() {
+    fileData = new FormData();
+  }
```
lucashu1, line 26: @Raghav-Maheshwari I don't really know enough about `componentWillUnmount` to comment on this, but do you know if these lines here are necessary? I know `componentWillUnmount` gives some warnings with the new React version

packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
```diff
@@ -0,0 +1,65 @@
...
+  async changeHandler(event) {
+    await this.setState({ note: event.target.value });
+  }
+
+  async handleSubmit(event) {
+    console.log('submitted');
+    console.log(`the item is ${JSON.stringify(itemId.itemId)}`);
+    event.preventDefault();
+    fileData.append('photo', this.fileInput.current.files[0]);
+    fileData.append('itemId', itemId.itemId);
```
lucashu1, line 37: See previous comment about doing `const { itemId } = this.props` within the `handleSubmit` function instead of using a global variable for `itemId`

packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx
```diff
@@ -0,0 +1,56 @@
...
+import Router from 'next/router';
+
+let mode = '';
+let language = '';
+
+class RequestTypeform extends React.PureComponent {
+  static async getInitialProps(ctx) {
+    let requestTypeformLink = '';
+    language = ctx.query.lang || 'en';
+    console.log(`langauge: ${language}`);
```
Comment (lucashu1) [inline packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx:13]: Let's comment out these `console.log`s from the frontend code unless they're necessary. Don't want potentially mischievous users snooping on what's going on behind the scenes!

packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable guard-for-in */
+import React from 'react';
+
+const FormData = require('form-data');
```
lucashu1, line 4: See other comment about trying to use `import` instead of `require` when possible

packages/backend/src/server/controller/items.js
```diff
@@ -5,6 +5,11 @@ import storeHelpers from "../util/storeHelpers.js";
 import sendgridHelpers from "../util/sendgridHelpers.js";
 import matchingHelpers from "../util/matchingHelpers.js";
 import errorHandler from "../util/errorHandler.js";
+import s3Helpers from '../util/s3Helpers';
+const multiparty = require('multiparty');
+
+let itemId = '';
+let note = '';
```
lucashu1, line 12: Avoid using global variables when possible. Let's keep these variables inside the `uploadConfirmationPhoto` function, if that's primarily where they'll be used

packages/backend/src/server/controller/items.js
```diff
@@ -241,9 +246,30 @@ async function updateItemDonorMessage(req, res) {
...
+  }
+  // eslint-disable-next-line no-else-return
+  else {
+    form.parse(req, function(err, fields, files) {
+      itemId = fields.itemId[0];
+      note = fields.note[0];
+      itemHelpers.updateConfirmationMessage(itemId, note);
+      Object.keys(files).forEach(function(fileKey) {
+        s3Helpers.uploadConfirmationPhotoToS3(itemId, files[fileKey]);
+      });
```
lucashu1, line 262: Will there be multiple files for a single itemId? Or can we just get the first element from `Object.keys(files)` and upload the corresponding file?

If we only want the first file, we can easily add an `await` statement as well:
`await s3Helpers.uploadConfirmationPhotoToS3(itemId, files[fileKey]);`

That way, we can know for sure that the photo upload succeeded before sending the user back a `200` response code

packages/backend/src/server/controller/items.js
```diff
@@ -241,9 +246,30 @@ async function updateItemDonorMessage(req, res) {
...
+  const form = new multiparty.Form();
+  if(req===undefined){
+    return res.status(400).json({message: 'Failed to post'})    
+  }
+  // eslint-disable-next-line no-else-return
+  else {
+    form.parse(req, function(err, fields, files) {
+      itemId = fields.itemId[0];
+      note = fields.note[0];
+      itemHelpers.updateConfirmationMessage(itemId, note);
```
lucashu1, line 259: Let's add an `await` here so that we know that this call succeeds before sending a `200` response code

packages/backend/src/server/util/s3Helpers.js
```diff
@@ -68,7 +70,31 @@ async function uploadPriceTagImageToS3(itemId, priceTagImageUrl) {
...
+    console.log("Uploading image to s3...");
+    const data = await s3.upload({
+      Body: imageStream,
+      Key: process.env.AWS_S3_IMAGE_FOLDER + '/item-' + itemId + '-confirmation' + jpgExtension,
+      Bucket: originalPhotosBucket,
+      ACL: "public-read",
+      ContentType: jpgMimeType
+    }).promise();
+    const s3PhotoUrl = data.Location;
+    itemHelpers.updateConfirmationPhotoLink(itemId, s3PhotoUrl);
```
lucashu1, line 86: This could go either way, but it might be good to add an `await` here so that we can be sure that the URL is updated before exiting out of the function

packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { ReactTypeformEmbed } from 'react-typeform-embed';
+import cookies from 'next-cookies';
+import Router from 'next/router';
+
+let mode = '';
+let language = '';
```
Comment (lucashu1) [inline packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx:7]: Try to avoid using global variables if possible. Let's move these declarations into `getInitialProps`.

packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx
```diff
@@ -0,0 +1,56 @@
...
+    if (!loggedIn) {
+      Router.push({
+        pathname: `/beneficiary-portal`,
+      });
+    }
+  }
+
+  render() {
+    const { requestTypeformLink } = this.props;
+    return <ReactTypeformEmbed url={requestTypeformLink} />;
```
Comment (lucashu1) [inline packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx:52]: I know we didn't do this for the other LA Typeform embed, but now that I think of it, it might be good to add a `Layout` wrapper around the `ReactTypeformEmbed`, so that we can change the title of the page as it appears in the browser tab. See other examples of where we use `Layout` with `title=...`. For the title of this page (and `ogTitle`, which is used for dynamic link previews), we could do something like "Duet | Item Request Form", and omit the other properties

packages/backend/src/server/controller/items.js
```diff
@@ -241,9 +243,27 @@ async function updateItemDonorMessage(req, res) {
   }
 }
 
+function uploadConfirmationPhoto(req,res){
+  let itemId = '';
+  let note = '';
+  const form = new multiparty.Form();
+  if(req===undefined){
```
Raghav-Maheshwari, line 250: I think we could simplify this to:
```
if (!req) {
   ...
}
```
@lucashu1 can you confirm?

packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx
```diff
@@ -0,0 +1,62 @@
...
+    let requestTypeformLink = '';
+    language = ctx.query.lang || 'en';
+    const loggedIn = cookies(ctx).loggedInBeneficiary === ctx.query.username;
+    if (process.env.APP_ENV === 'production') {
+      mode = 'production';
+    } else {
+      mode = 'development';
+    }
+    if (mode === 'production') {
+      if (language === 'en') {
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx:20]: you could probably simplify this to:
```
if (process.env.APP_ENV === 'production') {
   ... // prod typeforms
} else {
   ... // sandbox typeforms
}
```
instead of creating a separate `mode` variable

packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
```diff
@@ -0,0 +1,61 @@
...
+    console.log(`!!!${JSON.stringify(itemId)}`);
+    return { itemId };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { note: '' };
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.fileInput = React.createRef();
+    this.changeHandler = this.changeHandler.bind(this);
```
Raghav-Maheshwari, line 20: If you use arrow functions, you don't have to worry about binding the functions in your constructor. i.e:
```
changeHandler = (event) => {
 ...
}
```
You can read more about arrow functions [here](https://medium.com/@joespinelli_6190/using-arrow-functions-to-avoid-binding-this-in-react-5d7402eec64)

packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  async changeHandler(event) {
+    await this.setState({ note: event.target.value });
+  }
+
+  async handleSubmit(event) {
+    const { itemId } = this.props;
+    console.log(`item id is: ${itemId}`);
```
Raghav-Maheshwari, line 29: remove console log

packages/backend/src/server/controller/items.js
lucashu1: Yeah I think that would work!

Also, @OStouraiti let's wrap the entire `uploadConfirmationPhoto()` function in a try/catch block, as is done in the other controller functions. Send a `500` response if there's an error.

lucashu1:
@Raghav-Maheshwari FYI: the `ReactTypeformEmbed` package is showing an error in `dev` mode, but it doesn't actually affect the functionality at all. If you do `npm run start` instead of `npm run dev` (which I guess tells Next/React not to show those error popus), then everything works just fine, and I'm able to request items via the embedded Typeform.

You'll notice that https://giveduet.org/la-signup shows no errors, but if you navigate to that same page in `development` mode, the error does actually show up. So I think we can safely ignore those Typeform errors for now.

There's issue open issue in the ReactTypeformEmbed repo asking about this same thing: https://github.com/alexgarces/react-typeform-embed/issues/37

Screen Shot 2020-06-24 at 9 20 26 PM

packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
OStouraiti: It seems that I can't use async with an arrow function.

packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx
```diff
@@ -6,25 +6,19 @@ import Layout from '../../../components/shared/layout';
...
-    if (mode === 'production') {
       if (language === 'en') {
         requestTypeformLink = 'https://duet1.typeform.com/to/lS6WRw';
       } else if (language === 'fa') {
         requestTypeformLink = 'https://duet1.typeform.com/to/UOaMkY';
       } else if (language === 'ar') {
         requestTypeformLink = 'https://duet1.typeform.com/to/OwUmwF';
       }
-    } else if (mode === 'development') {
+    } else if ( process.env.APP_ENV != 'production') {
```
Comment (lucashu1) [inline packages/frontend/pages/beneficiary-portal/[username]/item-request.jsx:21]: You can leave this as just `else`. Also, note that `!=` and `!==` have different behavior in javascript! We'd probably want `!==`

packages/frontend/pages/beneficiary-portal/uploadConfirmationPhoto.jsx
lucashu1: Try `changeHandler = async (event) => { ... }`

https://stackoverflow.com/questions/42964102/syntax-for-async-arrow-function

Raghav-Maheshwari:
@lucashu1 @OStouraiti I think we can now close this in favor of #26

PR #20, opened 2020-06-24
Added link to family profiles on checkout page

nicolengsy:
As mentioned to @lucashu1 and @Raghav-Maheshwari, I felt that it would be helpful to have a direct link to specific family profiles on the checkout page, so that donors can go back to that family to add more items if wanted. Otherwise, there is currently no way to find a specific family except to be randomly taken there by the "Help a Family" button.


Raghav-Maheshwari:
🔥 🔥

PR #21, opened 2020-06-24
Send pickup message to donor and/or honoree every 24 hours

lucashu1:
Currently, pickup emails for each individual item, which can get annoying if you donated a lot of items at once. This PR gets the items that were picked up within the past 24 hours and sends pickup emails to each donor (and each honoree, if applicable).

**Changes:**
* Add `donor` to `item` data model
* Add `honoreeInfo` to `item` data model
* Add `donor_pickup_email_sent` flag to `items` table
* New `itemHelpers` functions: `getItemsNeedingDonorPickupEmail`, `setDonorPickupEmailSent`
* New env var: `CRON_INTERVAL_DONOR_PICKUP_EMAILS=0 16 * * *`
* New cron job: `sendDonorPickupEmails` (runs every day at 4pm Pacific)

**Todo:**
* Test out the whole flow in a local environment
* Set `donor_pickup_email_sent` to `1` for old items in production, so that donors don't get flooded with emails once this cron job runs for the first time

**Email Screenshots:**

Screen Shot 2020-06-24 at 3 30 18 PM
Screen Shot 2020-06-24 at 3 30 25 PM


packages/backend/src/server/cronJobs/donorsCronFunctions.js
```diff
@@ -0,0 +1,57 @@
...
+async function sendDonorPickupEmails() {
+  try {
+    // get items needing donor pickup email
+    const items = await itemHelpers.getItemsNeedingDonorPickupEmail();
+    if (!items.length) {
+      return;
+    }
+    // group items by donorEmail
+    const donorEmailToItems = new Map();
+    items.forEach(item => {
```
Raghav-Maheshwari, line 15: does this need to be async by any chance?

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -35,17 +35,30 @@ function sqlRowToItemObj(row) {
...
     requestedTimestamp: row.requested_timestamp,
     listedTimestamp: row.listed_timestamp,
     verifiedTimestamp: row.verified_timestamp,
     donationTimestamp: row.donation_timestamp,
     storePaymentInitiatedTimestamp: row.store_payment_initiated_timestamp,
     readyForPickupTimestamp: row.ready_for_pickup_timestamp,
-    pickedUpTimestamp: row.picked_up_timestamp
+    pickedUpTimestamp: row.picked_up_timestamp,
+    // TODO: deprecate these attributes
+    donorFirst: row.donor_first,
```
Raghav-Maheshwari, line 58: commenting as fyi - when these are deprecated, the "donated by x donor" will have to be updated in the front-end `itemCard` to reference `donor.donorFirst`

packages/backend/src/server/cronJobs/donorsCronFunctions.js
lucashu1: If it was an API route, we'd probably want to `await`, but since it's a cron job, nothing is really "awaiting" this. So I think we can keep is non-awaited for now

packages/backend/src/server/util/itemHelpers.js
lucashu1: Sounds like a future refactoring PR :)

packages/backend/src/server/util/sendgridHelpers.js
```diff
@@ -443,6 +451,57 @@ async function sendItemPickedUpEmail(itemId) {
...
+    let recipientList;
+    const templateId = "d-93bddeec5d0a4ccdb4609acb0ce3396f";
+    if (process.env.APP_ENV === "production") {
+      recipientList = [donorEmail, "duet.giving@gmail.com"];
+    } else {
+      recipientList = ["duet.giving@gmail.com"];
+      subjectTag = "[SANDBOX] ";
+    }
+    // group items into list of beneficiaries with "needs" arrays
+    const beneficiaries = groupItemsByBeneficiary(items);
```
Raghav-Maheshwari, line 470: does this need to be awaited?

lucashu1: No, it uses the info from the `item` objects themselves, so no `async` calls necessary

packages/backend/src/server/cronJobs/donorsCronFunctions.js
Raghav-Maheshwari: I guess I was just wondering if `donorEmailsToItems.forEach()` would ever run with an empty map

lucashu1: Ah gotcha, I think it'd just do nothing

PR #23, opened 2020-06-25
Add donors row for new donors

CodeBubble:
`insertDonorIntoDB()` handles checking if there is a new donor and adding a new row for each donation


packages/backend/src/server/util/donationHelpers.js
```diff
@@ -217,12 +217,66 @@ async function insertDonationIntoDB({
...
       `Successfully entered donation into DB: donationId: ${donationId}`
     );
+    insertDonorIntoDB({donorInfo})
     return donationId;
   } catch (err) {
     errorHandler.handleError(err, "donationHelpers/insertDonationIntoDB");
     throw err;
   }
 }
+async function insertDonorIntoDB({
```
lucashu1, line 227: Let's move this function into `donorHelpers`. And while `INSERT IGNORE` is probably ok for a 1-time MySQL Workbench query, we probably don't want to be relying on it in our actual codebase. Let's explicitly check if the `donor_email` exists before inserting the donor into the `donors` table.

See my other comment about creating a `getOrCreateDonorId()` helper function within `donorHelpers.js`.

To check if the `donor_email` already exists, you can use `donorHelpers.getDonorObjFromDonorEmail`, and check if the result is `null` or not.

While we're at it, let's change this line in `getDonorObjFromDonorEmail`:

```
if (results.length === 0) {
  throw new Error(`getDonorObjFromDonorEmail: donor not found for email: ${donorEmail}`);
}
```

to:

```
if (results.length === 0) {
  return null;
}
```

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -217,12 +217,66 @@ async function insertDonationIntoDB({
...
+    console.log(
+      `Successfully entered donor into DB: donorId: ${donorId}`
+    );
+    return donorId;
+  } catch (err) {
+    errorHandler.handleError(err, "donationHelpers/insertDonorIntoDB");
+    throw err;
+  }
+}
+//Not needed if INSERT IGNORE WORKS
```
lucashu1, line 260: See above comment about using (a modified version of) `donorHelpers.getDonorObjFromDonorEmail`. We can delete this commented code

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -217,12 +217,66 @@ async function insertDonationIntoDB({
     console.log(
       `Successfully entered donation into DB: donationId: ${donationId}`
     );
+    insertDonorIntoDB({donorInfo})
```
lucashu1, line 220: Remember that we'll need to log the `donor_id` inside the new `donation` row! So you'll have to call `insertDonorIntoDB({donorInfo})` _first_, before inserting the `donation` row into the database. And you'll have to modify the `INSERT INTO donations` statement to accept the `donor_id` instead of `donor_email`, `donor_fname`, `donor_lname`, and `donor_country` (since that info will already be in the `donors` table).

If you want, you can create a helper function in `donorHelpers.js` called something like `getOrCreateDonorId({ donorInfo })`. This function would:
1. check if a donor with that `donorEmail` exists, and if so, return that `donorId` 
2. if not, it will insert the new donor into the `donors` table, and return the newly inserted `donorId`

That way, we can just use 1 helper function to get a `donorId` to use in the `INSERT INTO donations` statement, regardless of whether that `donorEmail` already exists or not. See `stripeHelpers.getOrCreateStripePlanIdForAmountUsd()` for an example of this pattern in use.

Also, as a side note, let's modify the SQL statement in `getDonationObjFromDonationId` to query from `donations_view` instead of `donations`, since we will no longer be directly storing `donor_email`, `donor_fname`, etc. in the `donors` table.

`donations_view` _does_ contain additional info on the donor, because if you look at how `donations_view` is generated, it uses a `join` on `donations` and `donors`.

(By now, you've probably noticed that we use underscores and camelcase in the API server... it is what it is)

packages/backend/src/server/util/donorHelpers.js
```diff
@@ -29,17 +30,82 @@ async function getAllDonorEmails() {
     throw err;
   }
 }
+async function getOrCreateDonorId({
+  donorInfo
+}) {
+  console.log("inside getOrCreateDonorId")
```
lucashu1, line 36: Remove console.log

packages/backend/src/server/util/donorHelpers.js
```diff
@@ -29,17 +30,82 @@ async function getAllDonorEmails() {
...
+
+  const { email} = donorInfo;
+  try {
+    const donorObject = await getDonorObjFromDonorEmail(email)
+    if (donorObject) {
+      console.log("this aint new")
+      console.log(donorObject.donorId)
```
lucashu1, line 43: Remove console.logs

packages/backend/src/server/util/donorHelpers.js
```diff
@@ -29,17 +30,82 @@ async function getAllDonorEmails() {
...
+}
+async function insertDonorIntoDB({
+  donorInfo
+}) {
+  console.log(
+    `entered into insertDonorIntoDB`
+  );
+  console.log(
+    donorInfo
+  );
```
lucashu1, line 65: Remove console.logs

packages/backend/src/server/util/donorHelpers.js
```diff
@@ -29,17 +30,82 @@ async function getAllDonorEmails() {
...
-      "SELECT * from donors_view WHERE donor_email=?",
+      "SELECT * from donors WHERE donor_email=?",
       [donorEmail]
     );
     if (results.length === 0) {
-      throw new Error(`getDonorObjFromDonorEmail: donor not found for email: ${donorEmail}`);
+      return null;
     }
+    console.log(results[0]);
+    console.log(sqlRowToDonorObj(results[0]));
```
lucashu1, line 108: Remove console.logs

lucashu1:
Screen Shot 2020-06-26 at 6 33 20 PM

lucashu1:
@CodeBubble @Raghav-Maheshwari The above steps have been done, and this feature has been redeployed to prod! Congrats and thanks again @CodeBubble

PR #30, opened 2020-07-01
Passport Auth

lucashu1:
Use `passport` auth with sessions/cookies for stores and beneficiaries.

**Summary of changes:**
* Update `/login` routes: Use sessions to authenticate users, rather than hardcoded cookies
* Store session info in `sessions` table in MySQL (handled automatically by `express-mysql-session` package)
* **New environment variable:** `EXPRESS_SESSION_SECRET` (so pull from the .env files again!)
* Create a `checkLogin` route that uses the current sessionID to confirm logins within the beneficiary-portal/stores-portal. Specifically, `/api/beneficiaries/:username/checkLogin` and `/api/stores/:storeId/checkLogin`. Use `req.user` (populated by passport's middleware) to make sure usernames/ids match up.
* Redirect to login page if `checkLogin` call fails

**Packages:**
* https://www.npmjs.com/package/express-session
* http://www.passportjs.org/packages/passport-local/
* https://www.npmjs.com/package/express-mysql-session

**Tutorials:**
* https://medium.com/@evangow/server-authentication-basics-express-sessions-passport-and-curl-359b7456003d


packages/backend/src/server/controller/stores.js
```diff
@@ -1,63 +1,65 @@
...
+  // note: req.user populated by passport middleware (based on session ID)
+  if (req.user && req.user.storeId === Number(storeId)) {
+    return res.status(200).send({ message: `Logged in as store ${storeId}` });
   }
+  return res.sendStatus(401);
 }
 
+// async function sendBankTransfer(req, res) {
+//   try {
+//     let transferId;
```
Raghav-Maheshwari, line 31: are we deprecating this function?

lucashu1: It was just a test API route that I was using when developing the Transferwise code. I'd like to keep it there for now in case we need to test bank transfers in the future. I know they actually just sent us an email saying we have to upgrade the Transferwise API calls to v3 by October 1st (there are some breaking changes), so I'll likely be using it to test that migration over the coming weeks

packages/backend/src/server/controller/stores.js
```diff
@@ -1,63 +1,65 @@
...
-    );
-    res.status(200).send();
-  } catch (err) {
-    errorHandler.handleError(err, 'stores/testSendBankTransfer');
-    res.status(500).send();
+async function checkLogin(req, res) {
+  const { storeId } = req.params;
+  // note: req.user populated by passport middleware (based on session ID)
+  if (req.user && req.user.storeId === Number(storeId)) {
+    return res.status(200).send({ message: `Logged in as store ${storeId}` });
```
Raghav-Maheshwari, line 24: I'm wondering if it would be nicer to console log `Logged in as store ${storeName}` instead of `storeId` but not sure if there are any security implications

packages/backend/src/server/util/auth.js
```diff
@@ -1,28 +1,165 @@
...
-      password: duet_admin_password,
-    };
-    return done(null, admin_user, { scope: 'all' });
+    return done(null, adminUser, { scope: 'all' });
+  })
+);
+
+// Admin (local, uses sessions)
+passport.use(
+  'admin-local',
```
Raghav-Maheshwari, line 61: what does the 'local' tag mean?

packages/backend/src/server/util/auth.js
```diff
@@ -1,28 +1,165 @@
...
   })
 );
 
+// Stores
+passport.use(
+  'store-local',
+  new LocalStrategy(
+    {
+      usernameField: 'email',
+      passwordField: 'email',
```
Raghav-Maheshwari, line 79: sorry for all the questions: here the `usernameField` and `passwordField` are 'email', whereas for beneficiaries it's 'password'. Why is that?

packages/frontend/pages/stores-portal/[storeId].jsx
```diff
@@ -29,7 +44,11 @@ class StorePortal extends React.PureComponent {
   }
 
   render() {
-    const { items } = this.props;
+    const { loggedIn, items } = this.props;
+
+    if (!loggedIn) {
+      return null;
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/stores-portal/[storeId].jsx:50]: just confirming that this is where the redirect is happening if the user is not logged in?

packages/backend/src/server/util/auth.js
lucashu1: Correspond's to the `passport-local` strategy (http://www.passportjs.org/packages/passport-local/), which just means username + password based login. Not really sure why they call it `local` though. We don't use `local-storage` or anything

packages/backend/src/server/util/auth.js
lucashu1: This is how we tell `passport` which fields from the `POST` request `req.body` to use for authentication in the `.../login` routes. In the beneficiary login route, we use the `passcode` field, but in the stores login route, we use the `email` field

packages/frontend/pages/stores-portal/[storeId].jsx
Comment (lucashu1) [reply packages/frontend/pages/stores-portal/[storeId].jsx:50]: That's handled in `componentDidMount`:

```
if (!loggedIn) {
  Router.push({
    pathname: `/stores-portal`,
  });
}
```

packages/backend/src/server/controller/stores.js
lucashu1: Sure, that shouldn't be a security concern I think. Will do

lucashu1:
I think that would be a separate backlog item @Raghav-Maheshwari. Let's this in first so that @CodeBubble has an easier time merging with `master` (for his donor accounts branch), and then we can consider creating a new PR for those redirects later on

lucashu1:
@Raghav-Maheshwari Redirecting from `stores_portal/{storeId}` to `stores-portal/{storeId}` should work now! (Same for `beneficiary_portal`.) Give it a try on your local machine and feel free to merge this in if it works for you.

Cleaned up an unnecessary `fetch` request while I was at it: when using `giveduet.org/{itemId}`, it turns out that the `item` itself already contains all the info necessary for the redirect, so we can cut out the request to `/api/beneficiaries`.

PR #33, opened 2020-07-06
Fix cart tooltip visual bugs

nicolengsy:
Fixed visual bug for two-line item names, which previously made the content spacing look bad.

Also cleaned up some code from related files while I was at it!

Components changed: `CartToolTip`, `NavBar`


lucashu1:
@nicolengsy Make sure you run all the linting/prettier stuff so that it passes the GitHub lint check!

lucashu1:
Launched a review app here: https://duet-mono-website-pr-33.herokuapp.com

For some reason, the modal appears to be transparent the first time I open up a new window and add an item to cart, but the bug disappears for subsequent cart-adds. If you open up an incognito window, you might get the same bug.

Screen Shot 2020-07-06 at 1 24 15 PM

lucashu1:
Looks like some longer item names also still poke out from the modal:

This one is on https://duet-mono-website-pr-33.herokuapp.com/give/rezae5621

Screen Shot 2020-07-06 at 1 39 06 PM

But it looks much better now for 2-line item names! E.g.:

Screen Shot 2020-07-06 at 1 40 33 PM

Raghav-Maheshwari:
> Launched a review app here: https://duet-mono-website-pr-33.herokuapp.com
> 
> For some reason, the modal appears to be transparent the first time I open up a new window and add an item to cart, but the bug disappears for subsequent cart-adds. If you open up an incognito window, you might get the same bug.
> 
> Screen Shot 2020-07-06 at 1 24 15 PM
> 
> Edit: for some reason, I'm having a hard time replicating this bug now! Good thing I was able to get a screenshot... lmk if anyone has an idea of why this could be happening

@lucashu1 I think this might be happening because the cart modal is essentially an image with text overlayed on top of it. It's possible that for that specific page visit the connection to the CDN (where the image lives) was broken and so the image did not render (hence making the tooltip look transparent)

packages/frontend/components/shared/NavBar/CartToolTip.jsx
```diff
@@ -127,36 +123,36 @@ class CartToolTip extends React.PureComponent {
...
-                      alt="x"
-                      className={classes.removeItemIcon}
-                    />
-                    <span className={classes.removeLink}>Remove</span>
-                  </div>
+            <div className={classes.itemInfo}>
+              <img
+                src={item.image}
+                alt="item-img"
+                className={classes.itemImg}
```
Raghav-Maheshwari, line 130: image

Notice how the item image is kind of squashed? That's because the item is being resized by css `width` and `height` weirdly. 

We created a component called `<ResizedImage>` that basically takes in an image and its required width and height and re-uploads it to our CDN with the resized dimensions (you can find an example in `itemCard.jsx`)

Maybe you can try using the `<ResizedImage>` component here and see if it makes the item photos look better in the tooltip?

packages/frontend/components/Campaigns/LACovid19/checkout.jsx
```diff
@@ -238,7 +238,7 @@ class LACovid19Checkout extends React.PureComponent {
         </div>
         <div>
           {componentMounted && (
-            <React.Fragment>
+            <>
```
Raghav-Maheshwari, line 241: I like having the element specifically defined as `<React.Fragment>` (I just feel its more readable than `<>`), but curious to hear @lucashu1's thoughts about this too

lucashu1: I honestly didn't know `<>` was a thing (lol) so yeah I think `<React.Fragment>` might be more readable

nicolengsy: Interesting I didn't know about `<>` either... not sure how it made it there. I'll change it!

Raghav-Maheshwari:
@nicolengsy, unfortunately, I am seeing one more UI bug. I think when using the `ResizedImage` component, there's a delay in rendering the item's image when its added to the cart: 

Screen-Recording-2020-07-12-at-6 04 31-PM

This is probably because `<ResizedImage>` makes a new request on the client-side to AWS to run a lambda function and return a resized image only once the component is in frame. Maybe you could make a call to `getResizedImageUrl` in the utils lib the moment you're passing down the added item object into the cart tool tip?

Also, instead of using https://cdn.giveduet.org/website-images/cart_tooltip.png maybe we should store the image locally under the `public/static` folder so that we don't have to make an additional request to our CDN to render the tooltip background image.

nicolengsy:
@Raghav-Maheshwari I am now calling `getResizedImageUrl` in the NavBar, and using a regular `img` tag in CartToolTip instead. There's still some delay in rendering the image, but I can't quite tell if it's just my laptop (rendering has always been delayed for me). Please let me know how it looks from your end!

Raghav-Maheshwari:
@nicolengsy there's a delay in rendering on my end too, but I just realized that the same delay is occurring on the master branch too... so there might not be a way to get around this (@lucashu1 maybe you have an idea?)

Either way, this looks good to me now, so I will go ahead and approve the PR

lucashu1:
I hate to be a buzzkill, but now that more item names have been changed, we've started to encounter some 3-line item names, which still end up giving the modal a hard time :(

Unfortunately I don't have time to look into this today/tomorrow, but if you think that this is better than what's currently on the site @Raghav-Maheshwari, I'll trust your judgement if you want to merge this in first and push another change for 3-line item names later. If you'd rather have a full fix first, then that's definitely a good option too.

(this particular example is from http://localhost:3000/give/morad9741)

Screen Shot 2020-07-16 at 10 40 39 PM

Raghav-Maheshwari:
> I hate to be a buzzkill, but now that more item names have been changed, we've started to encounter some 3-line item names, which still end up giving the modal a hard time :(
> 
> Unfortunately I don't have time to look into this today/tomorrow, but if you think that this is better than what's currently on the site @Raghav-Maheshwari, I'll trust your judgement if you want to merge this in first and push another change for 3-line item names later. If you'd rather have a full fix first, then that's definitely a good option too.
> 
> (this particular example is from http://localhost:3000/give/morad9741)
> 
> Screen Shot 2020-07-16 at 10 40 39 PM

@lucashu1 Sorry I realized I never responded to this. 
My feeling is that if we merge this in now, pushing a fix for 3-line items might get put on the back burner. Let's aim to add a full-fix in this PR.

The reason why the design is breaking for longer name items is that we are using an image (with fixed proportions) to render the cart background. A solution to this would be to render the cart background using divs, and using pseudo-elements to render the top border with the "triangle". Here's an example: https://www.30secondsofcode.org/css/s/border-with-top-triangle

@nicolengsy do you think you'd be able to look into this? Otherwise I can also put this on my docket for this week!

nicolengsy:
image

Done! However, I couldn't quite get that box shadow on the triangle, so I used a larger opaque triangle behind the white triangle. It's also not rounded at the tip and on the sides, so it definitely doesn't look quite as good. Additionally, since the names are getting more descriptive and thus longer, I wonder if it would look better to make the tooltip wider, since Spence was likely designing for shorter names in mind, such as "Pan". 

Adding a rounded tip could help a lot visually, but I'm not sure if I can do it with some of the tricks on the internet, since I'm already using `:before` and `:after` to make triangles: (https://stackoverflow.com/questions/14446677/how-to-make-3-corner-rounded-triangle-in-css)

So this is as far as I got. If we want to make it better, I imagine we can have Spence give us the image of just his rounded triangle, and just place it above a rectangular container for the same effect.

lucashu1:
This looks great imo! I'm definitely not a CSS master, so @Raghav-Maheshwari if you think this is good to go, I think we can merge it in. If you have any tricks up your sleeve for those final tweaks, though, feel free to suggest them. Otherwise, I think we can live with a non-rounded triangle :)

Raghav-Maheshwari:
@nicolengsy I see the predicament! Yeah I'm not sure if we could get the triangle rounded with pure css... Let's try getting an svg image of the triangle from figma/spence and give that a try! Every pixel matters, and I have a feeling Spence wouldn't give they pointy triangle a pass in a VQA unfortunately...

Would you be able to coordinate with Spence in terms of getting an svg and uploading it to aws?

nicolengsy:
@lucashu1 @Raghav-Maheshwari Used the triangle SVG and it's looking good on my end, no lags! Let me know if it works the same for you. :)

Raghav-Maheshwari:
@nicolengsy yep the triangle looks great on my end too :)
Noticed that it needs to be off-centered when viewing on mobile!

Screenshot_20200727-122922

lucashu1:
Looks great to me @nicolengsy! (On mobile as well, with the new commit)

lucashu1:
@Raghav-Maheshwari Feel free to merge this in if you think it's ready

PR #37, opened 2020-07-09
Os fb account linking

OStouraiti:
This pull request has made changes in order to:

- add Messenger Consumer Chat plugin to the beneficiary portal
- extract the ref and referer_uri from the request when a new message is sent by a new facebook connection
- store these values in the messages table in the db

Note:
1) fb_chat_plugin_ref is the beneficiary username which we can later use to link the PSID to the beneficiaries table
2) to test make sure that the frontend domain is "whitelisted" in the Duet facebook page under Settings, Advanced Messaging
and that the backend domain is the same as that in the callback url found under Messenger, Settings, Webhooks in "My Apps" -> "Duet  Pickup Notifications"


packages/backend/src/server/util/fbHelpers.js
```diff
@@ -17,14 +17,16 @@ async function insertMessageIntoDB(message) {
   const { recipient } = message;
   const { content } = message;
   const { attachments } = message;
+  const { ref } = message;
+  const { refererUri } = message;
```
lucashu1, line 21: Let's actually combine these:

```
const { recipient, content, attachments, ref, refererUri } = message;
```

(good opportunity to clean up some old code while we're at it)

packages/frontend/components/BeneficiaryPortal/index.jsx
```diff
@@ -238,82 +240,91 @@ class BeneficiaryPortalComponent extends React.Component {
...
+            subtitle={t('pickedUp.subtitle')}
+            items={recentlyPickedUp}
+            noItemsText={t('pickedUp.noItems')}
+            donated
+          />
+        </div>
+        <div>
+          <MessengerCustomerChat
+            pageId="1147626915428137"
+            appId="572779639912312"
```
lucashu1, line 323: Let's define these as `const` variables at the top of the page, and then in here, do:
```
pageId={pageId}
appId={appId}
```
Might help make things _slightly_ cleaner (I think this is an appropriate use of global variables in this case)

packages/frontend/components/BeneficiaryPortal/index.jsx
```diff
@@ -238,82 +240,91 @@ class BeneficiaryPortalComponent extends React.Component {
...
-            alt="Duet Logo"
-            className={classes.logo}
-          />
-          <div className={classes.beneficiaryName}>
-            Beneficiary Portal: {beneficiary.lastName}
-          </div>
-          {isLoading && (
-            <div className={classes.loading}>
-              <ReactLoading type="spin" color="#32CBED" />
+      <Layout>
```
lucashu1, line 243: @Raghav-Maheshwari feel free to chime in too, but I think `<Layout>` usually acts as a wrapper in the `pages` folder. I don't think we typically use `<Layout>` for components. So I think we can remove it from here.

Raghav-Maheshwari: Yep the `<Layout>` component should only be used in `pages`, so we can remove from here (since this component is already a child component of `<Layout>` in `pages/beneficiary-portal/[username]`

packages/frontend/components/BeneficiaryPortal/index.jsx
Raghav-Maheshwari: @lucashu1 @OStouraiti any chance these IDs should be kept private and therefore be passed in as environment variables?

packages/frontend/components/BeneficiaryPortal/index.jsx
```diff
@@ -238,82 +240,91 @@ class BeneficiaryPortalComponent extends React.Component {
...
+              </div>
+            )}
+            <div className={classes.languageSelect}>
+              <span style={{ marginRight: 10 }}>{t('chooseLanguage')}</span>
+              <FormControl>
+                <Select
+                  value={language}
+                  onChange={e => {
+                    this.handleSetLanguage(e);
+                  }}
```
Raghav-Maheshwari, line 266: I think you can re-write as: `onChange={e => this.handleSetLanguage(e)}`

Raghav-Maheshwari:
Also, @OStouraiti make sure to merge master back into your branch and fix any merge conflicts you might have!

packages/frontend/components/BeneficiaryPortal/index.jsx
lucashu1: Nah I think these are safe to push up @Raghav-Maheshwari https://stackoverflow.com/questions/35402800/is-it-safe-to-have-facebook-app-id-embedded-in-the-ui-javascript-code.

Secret key should definitely be kept secret, though (as it already is)

lucashu1:
Review app: https://duet-mono-website-pr-37.herokuapp.com/beneficiary-portal

(Hooked up to another backend review app for this same PR)

PR #39, opened 2020-07-09
Builder CMS

Raghav-Maheshwari:
This PR sets up Builder on our website.

Current flow: 
visit /[...searchQuery] page

If the `searchQuery` slug matches with the builder page URL, then Builder will automatically render the builder page into our website.

_note:_ In the future we might want to think of renaming the [...searchQuery] route because right now it's acting like a catch-all route.

Also, we need to add a new client-side env variable in notion/heroku: `BUILDER_API_KEY`. The notion docs have been updated.


packages/frontend/pages/[...searchQuery].jsx
```diff
@@ -151,7 +170,25 @@ class Search extends React.PureComponent {
...
+      errorCode,
+      item,
+      redirect,
+      builderPage,
+      isItemId,
+    } = this.props;
+    // if this is a builder page
+    if (builderPage) {
+      return (
+        <div>
```
Comment (lucashu1) [inline packages/frontend/pages/[...searchQuery].jsx:184]: Is the reason for omitting `Layout` here because Builder creates its own text for descriptions, titles, etc.? Or would it still be worth it to include `Layout` so that we can control all the preview text and whatnot?

As long as Builder adds in its own page titles/descriptions and shows those cool-looking previews when we paste a link on Slack, then I guess we don't need `Layout`.

packages/frontend/pages/[...searchQuery].jsx
```diff
@@ -53,108 +53,104 @@ const redirects = {
...
+        // could not find item
+        console.log(`could not find item with itemId ${itemId}`);
+        const errorCode = 404;
+        return { props: { errorCode } };
+      }
+      if (item.status === 'REQUESTED' || item.status === 'LISTED') {
+        // item has not been verified (or is in graveyard)
+        console.log(`searchQuery: itemId ${itemId} is not yet verified`);
+        const errorCode = 404;
+        return { props: { errorCode } };
```
Comment (lucashu1) [inline packages/frontend/pages/[...searchQuery].jsx:122]: Make sure to remove the `props` layer if we're reverting to `getInitialProps`

PR #40, opened 2020-07-10
/api/beneficiaries supports username query param

lucashu1:
**Changes:**
* `/api/beneficiaries` supports `username` query param: e.g. `/api/beneficiaries?username={username1}&username={username2}` returns a list of 2 beneficiaries, `/api/beneficiaries?username={username1}` returns a list of 1 beneficiary
* OR use a comma-separated list: `/api/beneficiaries?username={username1},{username2}`
* `/api/beneficiaries` supports `withNeeds` and `withMeals` query params (default to `true`): e.g. `/api/beneficiaries?cause=refugee&withNeeds=false`
* `/api/items` supports a comma separated list: `/api/items?itemId={id1},{id2}`

Gonna go ahead and update the API docs now to reflect these changes ^

@CodeBubble This should be helpful for the Donor Impact page! (First, using the previous items-for-donors PR to get the items that a donor has donated, and then using this newly updated route to get the beneficiaries associated with those items)


Raghav-Maheshwari:
@lucashu1 wondering if designing the API route like:
`api/beneficiaries/{usernames}` where `{usernames}` can contain semicolon deliminated usernames is a bit more readable?

Inspiration from: https://api.stackexchange.com/docs/answers-by-ids

let me know what you think. Otherwise, this looks good to me!

lucashu1:
Oh huh, I actually haven't seen that pattern before @Raghav-Maheshwari 

What I've usually seen is `/api/{collection}/{id}` for returning a single objects, and `/api/{collection}?{query-params}` for returning lists (https://stackoverflow.com/a/9372278 <-- some of the other responses actually reference the link that you put as well). I guess the stackexchange semicolon-style could work too, though.

I'll actually go ahead and update this PR to accept a comma-separated list of IDs `/api/beneficiaries?username={username1},{username2}` as well, in case we want to use commas moving forward

PR #41, opened 2020-07-13
Connect Frontend and Backend Heroku Review Apps

Raghav-Maheshwari:
This PR connects frontend and backend review apps using dynamic env variables.

If `process.env.HEROKU_PR_NUMBER` exists, we find ourselves in a heroku review app. 

Set `DUET_API` and `DUET_WEBSITE` to 

`https://duet-mono-api-pr-${process.env.HEROKU_PR_NUMBER}.herokuapp.com` and 

`https://duet-mono-website-pr-${process.env.HEROKU_PR_NUMBER}.herokuapp.com` respectively. 

**When a new PR is made, create a review app for both the frontend and backend and they should point to each other**

**TODO:**
- Test main functionality and flows on [review app](https://duet-mono-website-pr-41.herokuapp.com/)
- Test main functionality on staging.giveduet.org to make sure that non-review apps still point to the right backend/frontend


lucashu1:
Did not know next.js config was a thing! That's cool

One thing that might be worrisome though is this (from https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration):

> A page that relies on publicRuntimeConfig must use getInitialProps to opt-out of Automatic Static Optimization. Runtime configuration won't be available to any page (or component in a page) without getInitialProps.

Funny enough though, the example they gave doesn't actually use `getInitialProps`... so I'm not really sure what's up with that. Regardless, if the documentation is true, then that might trip us up if we ever want to use the config on a static page for whatever reason.

This link suggests using the `env` config instead of `publicRuntimeConfig` for properties that won't change after the initial build: https://github.com/vercel/next.js/discussions/11493. It looks like your commit history says something about build-time envs, so maybe you already tried that and it didn't work? Was there some sort of issue when trying `env` or `process.env` with Heroku?

Raghav-Maheshwari:
@lucashu1 

Yeah I read that in the docs as well - that's why I tried using the `env` config instead. But what ended up happening is that you refer to the `env` config as `process.env.{}` and it seemed like it was still pulling in the variable from Heroku env variables instead of the `env` config.

I can't think of a reason that we'd use the runtime config on a static page, given that the runtime config is only defined for our API route and a static page, by definition, wouldn't be making any calls to our backend.

But if we do ever run into that use case, we can revisit this again!

lucashu1:
Gotcha @Raghav-Maheshwari. Just curious: did it not work when just using `process.env.DUET_API`?

And great point about this only being used for `DUET_API`, that's definitely a QED right there

Raghav-Maheshwari:
@lucashu1 yeah I tried `process.env.DUET_API` after setting the env config as so:

```
module.exports = {
  env: {
   DUET_API: process.env.HEROKU_PR_NUMBER
      ? `https://duet-mono-api-pr-${process.env.HEROKU_PR_NUMBER}.herokuapp.com`
      : process.env.DUET_API,
  },
}
```

But it was still making calls to `https://duet-mono-api-staging.herokuapp.com` which is what the Heroku `DUET_API` env variable is set to. So I feel like there might be some weird conflict thing going on there.

Will merge this in after doing some more testing!

lucashu1:
@Raghav-Maheshwari Gotcha. We could potentially remove `DUET_API` from the Heroku review app environment variables if we're only gonna be using both frontend + backend review apps together (and can always point them to each other). But otherwise, what we have in this PR should work too

Raghav-Maheshwari:
@lucashu1 Good point - for anything that's not a Heroku review app, we'll still revert to using Heroku's built-in env variables. Let me see if that works. It would simplify the code changes considerably!

Raghav-Maheshwari:
closing in favor of #44

PR #42, opened 2020-07-13
Donor Impact Page

CodeBubble:
http://localhost:3000/profile/[username] renders your impact page


packages/backend/src/server/util/donorHelpers.js
```diff
@@ -51,7 +52,26 @@ async function insertDonorIntoDB({ donorInfo }) {
     console.log(`Successfully entered donor into DB: donorId: ${donorId}`);
     return donorId;
   } catch (err) {
-    errorHandler.handleError(err, 'donationHelpers/insertDonorIntoDB');
+    errorHandler.handleError(err, 'donorHelpers/insertDonorIntoDB');
+    throw err;
+  }
+}
+
+async function insertDonorPasswordIntoDB(password, email) {
```
lucashu1, line 60: Let's use object-based arguments: `insertDonorPasswordIntoDB({ password, email })`

That way, when we call the function, we know the names of the arguments that we're passing in. Slightly more readable that way

packages/frontend/pages/donors-portal/[donorEmail]/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  return {
+    props: {
+      loggedIn,
+      donorEmail: query.donorEmail,
+      // items,
+    },
+  };
+}
+
+class StorePortal extends React.PureComponent {
```
Comment (lucashu1) [inline packages/frontend/pages/donors-portal/[donorEmail]/index.jsx:41]: *DonorPortal :)

packages/frontend/pages/donors-portal/[donorEmail]/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+    const { loggedIn } = this.props;
+
+    if (!loggedIn) {
+      Router.push({
+        pathname: `/donors-portal`,
+      });
+    }
+  }
+
+  render() {
```
Comment (lucashu1) [inline packages/frontend/pages/donors-portal/[donorEmail]/index.jsx:52]: If not loggedIn, then have `render` return `null`. That way, we won't accidentally reveal any info through the `render` function if it turns out that they're not actually logged in.

(I think `BeneficiaryPortal` in `master` does this if you wanna take a look)

packages/frontend/pages/donors-portal/index.jsx
```diff
@@ -0,0 +1,176 @@
...
+    fontSize: 30,
+    marginBottom: 20,
+  },
+  invalid: {
+    color: 'red',
+    marginBottom: 15,
+  },
+};
+
+class StoreLogin extends React.Component {
```
lucashu1, line 78: *DonorLogin

package.json
```diff
@@ -20,5 +20,8 @@
     "hooks": {
       "pre-commit": "npx lerna run --stream precommit"
     }
+  },
+  "dependencies": {
+    "bcrypt": "^5.0.0"
```
lucashu1, line 25: Move this to the` packages/backend` `package.json`! Reserve the root-level `package.json` for monorepo-management stuff like `lerna`

packages/backend/src/server/util/donorHelpers.js
```diff
@@ -16,6 +16,7 @@ function sqlRowToDonorObj(row) {
     donorFirst: capitalizeAndTrimName(row.donor_fname),
     donorLast: capitalizeAndTrimName(row.donor_lname),
     donorCountry: row.donor_country,
+    donorPassword: row.donor_password,
```
lucashu1, line 19: We probably don't want to be storing the password in the `donor` object, even if it is hashed. The safest thing to do is probably create a separate helper function that takes in a `donorEmail` (or username) and the attempted password, looks up the actual hashed password from the database, and simply compares them within the helper function. The function would return `true` if there's a match, or `false` if not.

That way, we don't reveal via the `donor` object what the hashed-password is.

CodeBubble:
new url: http://localhost:3000/profile/[username] renders your impact page

PR #52, opened 2020-07-23
Keyword-based search

lucashu1:
Support keyword-based search URLs: e.g. `giveduet.org/shoes`, `giveduet.org/jacket`, `giveduet.org/cooking-pan`. Should be helpful for future marketing pushes! (E.g. shoe drive)

**Backend:**
* `getItemObjs` helper supports `name` parameter (`items_view.item_name` must _contain_ `name` param)
* `/api/items/search` supports `name` query param: e.g. `/api/items/search?name=shoes&fairness=true`, or `/api/items/search?name=boy%27s%20shoes&fairness=true` (supports encoded URI components)

**Frontend:**
* Catch-all `searchQuery` page supports `searchTerms` dict for keyword-based search
* If `searchQuery` matches a keyword as defined in `searchTerms`, find an item that matches the search query, and redirect to that item's family page
* If `searchQuery` is not an item-id or a keyword, show a `404` error page

**Give it a try!** 
* https://duet-mono-website-pr-52.herokuapp.com/shoes
* https://duet-mono-website-pr-52.herokuapp.com/jacket
* https://duet-mono-website-pr-52.herokuapp.com/cooking-pan
* https://duet-mono-website-pr-52.herokuapp.com/kettle


packages/frontend/pages/[...searchQuery].jsx
```diff
@@ -48,6 +48,21 @@ const redirects = {
...
+  kettle: {
+    searchParams: 'name=kettle&status=verified&fairness=true',
+  },
+  jacket: {
+    searchParams: 'name=jacket&status=verified&fairness=true',
+  },
+  'cooking-pan': {
+    searchParams: 'name=cooking%20pan&status=verified&fairness=true',
+  },
+};
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/[...searchQuery].jsx:64]: Is there any way we could make the search more flexible?
I.e if down the road we wanted to be able to search for football/volleyball etc. like `giveduet.org/ball` could we do so without having to add another key-value pair to `searchTerms`?

packages/frontend/pages/[...searchQuery].jsx
```diff
@@ -138,13 +172,24 @@ class Search extends React.PureComponent {
       return null;
     }
     // item redirect: show loading screen while redirecting to family page
+    let title;
+    let ogTitle;
+    let description;
+    let image;
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/[...searchQuery].jsx:178]: let's define default values here rather than keeping them `null` or empty string, in case we are doing non-itemId based search

Comment (lucashu1) [reply packages/frontend/pages/[...searchQuery].jsx:178]: There already are default values for each of these in `Layout`

packages/frontend/pages/[...searchQuery].jsx
Comment (lucashu1) [reply packages/frontend/pages/[...searchQuery].jsx:64]: Good point! I like that idea. One option is to store these in a database table, but that means that for every `searchQuery` render, we'd need to make a call to the database. I wonder if it's possible to pre-load the key-value pairs  from the database when we first initialize the Next.js app? Possibly using the Next.js config settings. I can look into that tomorrow

lucashu1:
@Raghav-Maheshwari One idea to move things in a more flexible direction: I added a table in the for `search_keywords` table in the database, and created a new API route `/api/items/search/keywords` that fetches from that table, instead of using a hardcoded `searchTerms` object.

The problem with having fully flexible keywords is that technically, _anything_ could be used as a keyword. That means that we'd never really get a `404` error, and that we'd always try to find an item that matches any possible query string (even something like `giveduet.org/blah`). So for now, it might be useful to constrain the list of keywords. I think that using the database could be a workable middle ground without having to make any code changes every time we want to add a new keyword.

Optionally, instead of having a global variable to store the `searchKeywords`, we could force the Next.js server to fetch from `/api/items/search/keywords` every time (instead of caching it), so that we don't have to redeploy if we add a new keyword to the database. Right now, `searchKeywords` only gets loaded once, and doesn't get refreshed if we add something to the database.

Lmk if you have any other ideas on how to keep a list of keywords that's not _too_ flexible

Raghav-Maheshwari:
@lucashu1 You don't need to apologize about the wall of text; if anything, more detail is more helpful! 😄 

How important is it from the BizDev side to have item keywords be on `giveduet.org/{item-keyword}`?
If we were concerned about 404 errors (or a lack of them), we could always specify item search at `giveduet.org/items/{item-keyword}` which in my opinion is cleaner anyways and it separates logic out from `[...searchQuery].jsx`.

If we decided to go the `giveduet.org/items/{item-keyword}` route then I think we should have the back-end match on the word level (re: your edit). I'm also not a huge fan of storing key-words in a database and having the front-end query on them as it feels like more overhead than is necessary.

lmk your thoughts!

lucashu1:
@Raghav-Maheshwari What about using `giveduet.org/item/{itemId}` for specific items, and `giveduet.org/search/{searchQuery}` for name-based search?

If we were being fully RESTful, `giveduet.org/search?q={searchQuery}` would probably be more accurate, but `giveduet.org/search/{query}` is probably a lot simpler for non-engineering users to remember and use

Raghav-Maheshwari:
@lucashu1 awesome! Yes I like this solution too :)

packages/frontend/pages/item/[itemId].jsx
```diff
@@ -0,0 +1,113 @@
...
+
+class ItemRedirect extends React.PureComponent {
+  componentDidMount() {
+    // route to beneficiary page, with item preselected
+    const { item } = this.props;
+
+    if (item) {
+      const href = `/give/${item.beneficiaryUsername}?preselect=${item.itemId}`;
+      const as = href;
+      Router.push(href, as);
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/item/[itemId].jsx:74]: if we set `as` = `/give/${item.beneficiaryUsername}` for cleaner URLs, would it break the preselect behavior?

packages/frontend/pages/search/[searchQuery].jsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+class ItemSearch extends React.PureComponent {
+  componentDidMount() {
+    // route to beneficiary page, with item preselected
+    const { item } = this.props;
+
+    if (item) {
+      const href = `/give/${item.beneficiaryUsername}?preselect=${item.itemId}`;
+      const as = href;
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/search/[searchQuery].jsx:61]: same comment as above

packages/frontend/pages/item/[itemId].jsx
Comment (lucashu1) [reply packages/frontend/pages/item/[itemId].jsx:74]: I think it would @Raghav-Maheshwari.

Actually, looking at this (https://nextjs.org/docs/routing/introduction), I think we're currently using `href` wrong in our links/routing. It looks like `href` is supposed to be the actual page template-name as defined in our `/pages` folder (so in this case, `/give/[beneficiaryIdOrUsername]`), and `as` is supposed to be the actual dynamic link (`/give/${item.beneficiaryUsername}?preselect=${item.itemId}`).

Our current code doesn't seem to break anything though, so maybe we can add that to the backlog

PR #54, opened 2020-07-23
Fixed login bug for item request form

OStouraiti:
Adapted the way boolean loggedIn is computed by getting the server side props created by passport.
Replaced get InitialProps with getServerSideProps to have access to req and make the fetch call to the checkLogIn API.


lucashu1:
Great find @OStouraiti! Merging and deploying this now, since it's a pretty important bug fix

PR #61, opened 2020-08-03
Refugee-side subscriptions

nicolengsy:
Allocation of contributions is found in `donationHelpers./allocateContributions()` :
- While funds >= $25, donate items between ~~$20-25~~ $10-25
- While funds >= ~~$8~~ $10, donate items between ~~$5-8~~ $2-10
- Donate remaining funds to one selected $25+ item

Created various `itemHelpers.js` helper functions to aid the functionality: `getDonatableItemsWithCheckoutPrices()`, `updateCheckoutPricesInItemsObjs()`, which are used  This is because I need to calculate the remaining price in USD after making SQL calls.

Also updated `itemHelpers/sqlRowToObj()` to match updated data model.

~~NOTE: There is a dependency cycle that needs to be fixed, with imports going from `donationHelpers` -> `paypalHelpers` -> `sendGridHelpers` -> `donationHelpers`.~~

### Post-Handoff -> Lucas:

**Summary of additional changes since handoff:**

* Show partially funded items in donation tracker
* Integrate Sendgrid monthly donation confirmation email
* Set minimum item price to $0.25
* Clean up some SQL pipe-delimiter code (use a `splitSqlCol` helper function)
* Added some `utils` in `packages/frontend` to clean up currency-conversion code (more to be done in a future PR)
* Refactor our `donation.items` in favor of `donation.contributions`
* Show tooltip for subsidized items

Did some testing with all our major donation flows, and it seems like everything still works!

Note: Left the `item.donor` object in for now (it uses the donor info from the fully-funded contribution), because it's still used in a lot of places, and would be a big pain to refactor out right now

**Update:** Also tested the subscription webhook after setting the Stripe webhook to point to the backend review app. Looks like it's good to go.

**TODO when merging:**
* Create `contributions` table in prod database
* Re-populate `contributions` table (sandbox and prod)
* Replace `items_view` with `items_view_crowdfunding` (sandbox and prod)

### Screenshots:
**Email:**
Screen Shot 2020-08-16 at 5 39 49 PM

**Donation tracker:**
Screen Shot 2020-08-16 at 5 39 22 PM

**Subsidized price:**
Screen Shot 2020-08-24 at 11 20 07 AM

**Subsidized price tooltip:**
Screen Shot 2020-08-24 at 11 20 16 AM
(URL will get replaced with the current `baseUrl` depending on what environment we're in)


nicolengsy:
Added changes to relevant helper functions for price filter in `itemHelpers.js` so that it searches for items using the remaining price (USD) rather than the original item price (EUR).

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -281,6 +284,126 @@ async function removeDonationFromDB(donationId) {
   }
 }
 
+async function insertContributionIntoDB(
+  donationId,
+  itemId,
+  contributionAmtUsd,
+  isFullyFunded
+) {
```
lucashu1, line 292: Use object destructing parameters

packages/backend/src/server/controller/donations.js
```diff
@@ -131,7 +142,17 @@ async function captureTransaction(req, res) {
         return captureTransactionMethods.capturePayPalMealTransaction(req, res);
       }
       if (itemInfo) {
-        return captureTransactionMethods.capturePayPalItemTransaction(req, res);
+        // TODO: euros or usd??
```
lucashu1, line 145: Is this comment supposed to be removed?

packages/backend/src/server/controller/donations.js
```diff
@@ -97,7 +97,18 @@ async function captureTransaction(req, res) {
         return captureTransactionMethods.captureStripeMealTransaction(req, res);
       }
       if (itemInfo) {
-        return captureTransactionMethods.captureStripeItemTransaction(req, res);
+        const response = captureTransactionMethods.captureStripeItemTransaction(
+          req,
+          res
+        );
+
+        // TODO: get donation id ?? usd or euros?
```
lucashu1, line 105: Is this todo taken care of?

packages/backend/src/server/controller/donations.js
```diff
@@ -97,7 +97,18 @@ async function captureTransaction(req, res) {
...
+        await donationHelpers.insertContributionIntoDB(
+          response.donationId,
+          itemInfo,
+          amount
+        );
```
lucashu1, line 110: Don't do this here -- do it in the actual function (`captureTransactionMethods.captureStripeItemTransaction`) instead

packages/backend/src/server/controller/donations.js
```diff
@@ -131,7 +142,17 @@ async function captureTransaction(req, res) {
...
+        // TODO: euros or usd??
+        const response = captureTransactionMethods.capturePayPalItemTransaction(
+          req,
+          res
+        );
+        await donationHelpers.insertContributionIntoDB(
+          response.donationId,
+          itemInfo,
+          amount
+        );
```
lucashu1, line 154: Again, do this in `captureTransactionMethods.capturePayPalItemTransaction`

packages/frontend/components/FamilyProfile/ItemCard.jsx
```diff
@@ -240,9 +240,11 @@ const ItemCard = ({
 
   let donatedByText;
   if (need.donorFirst && need.donorLast) {
+    const donorsFirst = need.donorFirst.split('|');
+    const donorsCountry = need.donorCountry.split('|');
     donatedByText = (
       <div className={classes.donatedBy}>
-        Donated by {need.donorFirst} {flags[need.donorCountry]}
+        Donated by {donorsFirst[0]} {flags[donorsCountry[0]]}
```
lucashu1, line 247: This will be replaced by whichever donor had the fully-funded contribution

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -281,6 +284,126 @@ async function removeDonationFromDB(donationId) {
...
+      await insertContributionIntoDB(
+        donationId,
+        item.itemId,
+        item.amount,
+        item.isFullyFunded
+      );
+    })
+  );
+
+  console.log('donationHelpers/allocateContributions: allocated contributions');
```
lucashu1, line 404: A more detailed console log would be helpful! E.g. donationId, donation amount, and list of item IDs

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -469,6 +515,28 @@ function getNextItemStatus(oldStatus) {
   }
 }
 
+async function updateCheckoutPricesInItemsObjs(items) {
+  try {
+    await Promise.all(
+      items.map(async (item) => {
+        // eslint-disable-next-line no-param-reassign
```
lucashu1, line 522: To get around this, you can instead do something like `return { ...item, checkoutPriceUsd }` within the `map`, and then use a list variable to store the output of the `Promise.all`. That way, you'll be returning a new item instance with the `checkoutPriceUsd` added, instead of reassigning a parameter to the old object.

Also, super nitpicky, but maybe something like `priceUsd` is a more fitting variable name to use here? Mainly bc the items haven't actually been checked out yet. For now, we've only been using "checkout" within the context of items that have already been full-funded.

You could then rename the function something like `attachPriceUsdToItemObjs()`

packages/frontend/components/FamilyProfile/ItemCard.jsx
```diff
@@ -273,7 +275,7 @@ const ItemCard = ({
                   {`$${fx(need.price)
                     .from('EUR')
                     .to('USD')
-                    .toFixed(2)}`}
+                    .toFixed(2) - need.totalContributionUsd}`}
```
lucashu1, line 278: `.toFixed()` returns a string, so you might accidentally be subtracting a Number from a string

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -281,6 +284,126 @@ async function removeDonationFromDB(donationId) {
...
+  const itemsTier3 = [];
+  const selectedItems = [];
+
+  items.forEach((item) => {
+    const remainingPriceUsd = item.checkoutPriceUsd - item.totalContributionUsd;
+    if (remainingPriceUsd >= 20 && remainingPriceUsd <= 25) {
+      itemsTier1.push(item);
+    } else if (remainingPriceUsd >= 5 && remainingPriceUsd <= 8) {
+      itemsTier2.push(item);
+    } else if (remainingPriceUsd >= 25) {
```
Raghav-Maheshwari, line 338: since you check `remainingPriceUsd <= 25` in the first if statement you want `remainingPriceUSD > 25` in the last else if.

packages/frontend/components/FamilyProfile/ItemCard.jsx
Raghav-Maheshwari: ^^ This will always be the bane of my existence

lucashu1:
@Raghav-Maheshwari Just had a chat with @nicolengsy about this after our standup. I agree with you that at some point, we should probably move in the direction of accepting multiple currencies. I do think that making that full transition might end up being a lot more work than we can really predict right now, though. My main concern is that if we're going to expand to LA in full in 2021, then we'll need to refactor the entire `items` table to support multiple currencies, too -- and there will probably be a bunch of spillover changes that we'd have a pretty tough time trying to plan around right now. Realistically, I think that "multiple currency support" is probably a much larger project that we'll have to tackle after the 2020 holiday season is over.

To give us a headstart though, what we could do for this PR is simply to separate the `contribution_amt_usd` column in the `contributions` table (and `items_view` as well) into separate `contribution_amt` and `currency` columns -- which would be a pretty easy change to make. (For now, everything would just be in `USD`.) In the contribution-related data models, we'd also split `contributionAmtUsd` into `contributionAmt` and `contributionCurrency`.

As far as the rest of the backend logic goes (e.g. `allocateContributions()`), I think sticking with Option 1 would allow us to get this feature pushed out sooner rather than later, and give @nicolengsy more time to get a head start on the Donation History components, which would be a huge help for moving into the holiday season. (Plus, code changes are much easier to make than data model/database schema changes, so this should be relatively easy to change later down the line.)

@nicolengsy Lmk if that sounds good to you! (I know this differs slightly from what we spoke about during our call, but I just thought of this after the fact lol)

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -34,6 +34,29 @@ function sqlRowToDonationObj(donationRow, donorObj) {
...
+  for (let i = 0; i < Number(donationRow.num_donated_items); i += 1) {
+    contributions.push({
+      item: {
+        itemId: itemIds[i],
+        itemName: itemNames[i],
+        donorMessage: donorMessages ? donorMessages[i] : '',
+      },
+      contributionAmt: Number(contributionAmts[i]),
+      contributionCurrency: contributionCurrencies[i],
+      fullyFunded: fullyFunded[i],
```
lucashu1, line 56: I think you might have to manually cast this to a `Boolean`? Not sure tbh

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -34,6 +34,29 @@ function sqlRowToDonationObj(donationRow, donorObj) {
...
+  const donorMessages = donationRow.donor_message
+    ? donationRow.donor_message.split('|')
+    : '';
+  for (let i = 0; i < Number(donationRow.num_donated_items); i += 1) {
+    contributions.push({
+      item: {
+        itemId: itemIds[i],
+        itemName: itemNames[i],
+        donorMessage: donorMessages ? donorMessages[i] : '',
+      },
```
lucashu1, line 53: It might be better to have each of these be the full `item` object, instead of just ID+name+message. That way, we can make one call to `/api/donations/{donationId}` to be able to render all the relevant info for that donation (including the item image, family image, etc.).

One idea for how to do this would be to call `getItemObjsFromItemIds` to get the relevant `item` objects for this donation, and then to retrieve the individual item from the resulting list within the `for` loop

packages/frontend/components/FamilyProfile/ItemCard.jsx
```diff
@@ -272,10 +272,12 @@ const ItemCard = ({
...
-                  {`$${fx(need.price)
-                    .from('EUR')
-                    .to('USD')
-                    .toFixed(2) - need.totalContributionUsd}`}
+                  {`$${Number(
+                    fx(need.price)
+                      .from('EUR')
+                      .to('USD')
+                      .toFixed(2)
+                  ) - need.totalContributionUsd}`}
```
lucashu1, line 280: I think this should do the trick:

```
(fx(need.price).from('EUR').to('USD') - need.totalContributionUsd).toFixed(2)
```

We still want to have `.toFixed(2)` be on the very outside, so that the final result is a string with 2 decimal points. Otherwise, JS can do some weird stuff with omitting the decimal points when rendering if the final variable is left as a `Number`

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -5,7 +5,71 @@ import errorHandler from './errorHandler';
 import donorHelpers from './donorHelpers';
 import currencyHelpers from './currencyHelpers';
 
-function sqlRowToItemObj(row) {
+function sqlRowToItemObj(row, withDonations = false) {
```
lucashu1, line 8: Let's put the `withDonations` in an optional object-style argument:

`sqlRowToItemObj(row, { withDonations=false }={})`

It's typically best to have the `options` be in an object-style argument like that.

The `={}` means that the options argument is optional (it makes the argument an empty object by default)

packages/backend/src/server/controller/captureTransaction.js
```diff
@@ -512,14 +518,20 @@ async function capturePayPalItemTransaction(req, res) {
...
           await itemHelpers.updateCheckoutPriceUsd(
             priceInfo.itemId,
             priceInfo.priceUsd
           );
+          await donationHelpers.insertContributionIntoDB(
+            donationId,
+            priceInfo.itemId,
+            priceInfo.priceUsd,
+            1
+          );
```
lucashu1, line 534: Update this (and other calls) to use the object destructuring syntax

packages/backend/src/server/controller/captureTransaction.js
```diff
@@ -93,14 +93,20 @@ async function captureStripeItemTransaction(req, res) {
...
```
lucashu1, line 109: Update this to use the object destructuring syntax

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -281,6 +310,164 @@ async function removeDonationFromDB(donationId) {
...
+      `INSERT INTO contributions (
+        item_id, donation_id, contribution_amt, contribution_currency, fully_funded
+      ) 
+      VALUES (?,?,?,?,?)`,
+      [itemId, donationId, contributionAmt, contributionCurrency, isFullyFunded]
+    );
+
+    if (isFullyFunded) {
+      // send PayPal payout to stores with payment_method='paypal' (don't await)
+      paypalHelpers.sendNecessaryPayoutsForItemIds(itemId);
```
lucashu1, line 333: This should go elsewhere -- reserve `insertContributionIntoDB` for contributions-specific code. Otherwise, if we try to do too much in each helper function, we end up increasing our chances of circular dependencies.

My suggestion would be to move this `paypalHelpers` call into the webhook controller. You could have `allocateContributions()` just return the `selectedItems` objects (i.e., just selecting items instead of actually updating them), and then handle the actual updating of the `// insert async contributions into DB` (as well as sending any necessary payouts) in the webhook controller.

A suggestion for how to clean up the subscription webhook code while we're at it: Similarly to how we have separate `captureTransaction` controller functions for different donation flows (paypal, stripe, etc.), you could do something similar for the subscriptions webhooks (refugee-side or LA-side).

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -281,6 +310,164 @@ async function removeDonationFromDB(donationId) {
...
+      // update items table for legacy support
+      await conn.query(
+        `UPDATE items SET status=?, donation_id=? WHERE item_id=?`,
+        ['PAID', donationId, itemId]
+      );
+    }
+    console.log(
+      `Successfully inserted contribution ${results.insertId} into database`
+    );
+    return results;
```
lucashu1, line 344: Unless you need to return the full `results` object, it might be cleaner to just return the `insertId` (i.e. the `contribution` ID)

packages/backend/src/server/util/itemHelpers.js
nicolengsy: Thanks for the explanation!! Got confused with this notation

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -38,36 +103,45 @@ function sqlRowToItemObj(row) {
...
+    donorFirst: donations ? donations[donations.length - 1].donorFirst : null,
+    donorLast: donations ? donations[donations.length - 1].donor_last : null,
+    donorEmail: donations ? donations[donations.length - 1].donor_email : null,
+    donorCountry: donations
+      ? donations[donations.length - 1].donor_country
+      : null,
+    donationId: donations ? donations[donations.length - 1].donation_id : null,
+    donationTimestamp: donations
+      ? donations[donations.length - 1].donationTimestamp
+      : null,
```
lucashu1, line 124: Oh one more thing! Let's only set these attributes (`donorFirst`, `donorLast`, etc.) if the item has been fully funded. One way to do that would be to check if the status is either `PAID`, `READY_FOR_PICKUP`, or `PICKED_UP`, but maybe there's an easier way to do it using the donations/contributions directly (e.g. if at least one of the contributions associated with the item has fully_funded = 1)

nicolengsy:
@lucashu1 Updated the files with changes based on comments, and left the comment `TODO: implement multiple-currency flexibility` in `donationHelpers.js` and `itemHelpers.js`, as I calculated the remaining prices assuming all contributions are in USD.

Also, current allocation algorithm may need wider ranges of prices instead of [20,25] and [5,8], because the subscriptions easily wipe out all the items within those price ranges, and all the money goes to an item of $25+, and takes away the opportunity for donors to donate a complete item. Let me know what you think may be a more reasonable price range!

lucashu1:
Awesome, thank you @nicolengsy!

Re: price ranges, I trust you to find some reasonable ranges! You can refer back to the `items` table to see the distribution of prices historically. Or if you log into metabase and go into "Refugee Operations", you'll see some relevant metrics as well.

Will take a more thorough look through the code later, add in the Sendgrid confirmation email, etc.

PS it looks like there's a merge conflict you may need to take care of too

packages/backend/src/server/controller/donations.js
```diff
@@ -251,6 +252,47 @@ async function handleStripeWebhook(req, res) {
             });
           }
         } else {
+          // allocate contributions to refugee items
```
Raghav-Maheshwari, line 255: can we actually break out the cause specific bookkeeping into separate helper functions? Might help with readability

packages/backend/src/server/controller/donations.js
```diff
@@ -251,6 +252,47 @@ async function handleStripeWebhook(req, res) {
...
+          const selectedItems = await donationHelpers.allocateContributions(
+            donationId,
+            amount
+          );
+
+          // insert contributions into DB
+          const contributionIds = [];
+          const selectedItemIds = [];
+          const fullyFundedItemIds = [];
+          await Promise.all(
```
Raghav-Maheshwari, line 265: let's also put this functionality in a shared helper function in order to make it DRY (don't repeat yourself)

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -88,10 +114,16 @@ async function getDonationObjFromDonationId(donationId) {
...
-    const items = await getItemObjsForDonationId(donationId);
-    donationObj.items = items;
+    let donationObj = sqlRowToDonationObj(donationRow, donor);
+    donationObj = await attachContributionsToDonationObj(
+      donationRow,
+      donationObj
+    );
+
+    // get fully funded items associated with donation
+    donationObj.items = donationObj.contributions
```
Raghav-Maheshwari, line 124: maybe rename to `donationObj.compeletdItems` or similar to specify that these items were fully funded from this donation?

packages/backend/src/server/util/donationHelpers.js
```diff
@@ -123,19 +154,37 @@ async function getDonationObjsForDonorUsername(donorUsername) {
...
-    });
+    const donationObjs = await Promise.all(
+      donationRows.map(async (row) => {
+        // convert SQL row to donation object
+        const donor = donorHelpers.sqlRowToDonorObj(row);
+        let donationObj = sqlRowToDonationObj(row, donor);
+        // attach contributions
+        donationObj = await attachContributionsToDonationObj(row, donationObj);
+        // attach items, donatedMeals attributes
+        donationObj.items = donationObj.contributions
```
Raghav-Maheshwari, line 165: I feel like another way to do this would be to only track a donation object's items through `contributions` instead of separating it out into `donationObj.items` and `donationObj.contributions`. 

If we use `contributions` as our source of truth for the many-to-many relationship between donations and items, then we'll know which items were fully-funded on the item-level.

I just feel like that's cleaner than having to keep track of `.items` and `.contributions` (also b/c `.items` is a subset of `.contributions`?

what do you think @lucashu1

packages/backend/src/server/util/itemHelpers.js
```diff
@@ -184,17 +293,71 @@ async function getItemObjs(options = {}) {
...
+      items = await attachPriceUsdToItemObjs(items);
+
+      items = items.filter((item) => {
+        // calculate remaining price for each item
+        // TODO: implement multiple-currency flexibility
+        let totalContributionUsd = 0;
+        if (item.donations) {
+          item.donations.forEach((donation) => {
+            totalContributionUsd += donation.contributionAmt;
+          });
```
Raghav-Maheshwari, line 320: so I noticed you're doing this `forEach()` in a couple places to get the `totalContributionUsd` amount towards the item. Thoughts on just adding a property to the item object called `totalContributionUsd`? Might be able to save you from the `forEach((donation ... )` logic and will also make it easier on the front-end to display updated item prices to the donor

packages/frontend/components/FamilyProfile/ItemCard.jsx
```diff
@@ -239,14 +239,24 @@ const ItemCard = ({
...
     );
   }
 
+  // get total contribution
+  // TODO: implement multiple-currency flexibility
+  let totalContributionUsd = 0;
+  if (need.donations) {
+    need.donations.forEach(donation => {
+      totalContributionUsd += donation.contributionAmt;
+    });
```
Raghav-Maheshwari, line 257: ah there you go ;)

packages/frontend/components/FamilyProfile/ItemCard.jsx
```diff
@@ -270,10 +280,11 @@ const ItemCard = ({
...
                 <div className={classes.price}>
-                  {`$${fx(need.price)
-                    .from('EUR')
-                    .to('USD')
-                    .toFixed(2)}`}
+                  {`$${(
+                    fx(need.price)
+                      .from('EUR')
+                      .to('USD') - totalContributionUsd
+                  ).toFixed(2)}`}
```
Raghav-Maheshwari, line 287: so this is great! But when we calculate the total donation amount from the cart, are we using this updated "subsidized" price? I'm not sure we are, so you might want to add that to this PR.

packages/backend/src/server/util/donationHelpers.js
nicolengsy: I would rename it but I'm assuming this attribute is being used elsewhere in the code right now and renaming it would require refactoring it everywhere else! Is that true @lucashu1 ?

packages/backend/src/server/util/donationHelpers.js
nicolengsy: Oops this is relating to my comment on the previous comment, but I didn't take away the items attribute in case it is currently used elsewhere. Should I get rid of it?

lucashu1: That would definitely be cleaner -- I've honestly lost track of where else donation.items is currently used at the moment. Definitely in the donation tracker (giveduet.org/donation?donationId={id}), and I'm sure in a few other places too. @nicolengsy maybe you can poke around and do some digging?

If it's used in too many places right now (or just to err on the safe side regardless), we could keep `donation.items` for backward compatibility, and then add a `TODO` indicating that we want to refactor it out (in favor of `contributions`) at some point in the future.

But we should still replace as many instances of `donation.items` as possible with `donation.contributions`, to give us a head start on that refactoring

packages/backend/src/server/util/donationHelpers.js
lucashu1: Yeah, it would require refactoring. See my other comment too

packages/backend/src/server/controller/donations.js
nicolengsy: @Raghav-Maheshwari I just moved them but it creates new dependency cycle errors! Should I go ahead and ignore it or leave it in the controller?

packages/backend/src/server/util/donationHelpers.js
nicolengsy: Added the `TODO`! I'll poke around after I get more headstart on Manage Accounts

packages/backend/src/server/controller/donations.js
lucashu1: I don't think this should result in a dependency cycle error if the files have been set up correctly. See how the donations controller imports from captureTransaction (also in controllers folder), and see if you can replicate that

packages/backend/src/server/util/donationHelpers.js
lucashu1: Sounds like a good plan

packages/frontend/components/FamilyProfile/ItemCard.jsx
nicolengsy: Wait, what does this mean

Raghav-Maheshwari: oh in my comment above I said if we just added a new property `totalContributionUsd` to the item object we wouldn't have to do this `forEach` everywhere

nicolengsy:
Changes made:
- Minor updates from PR comments
- Updated cart tooltip and checkout to use subsidized prices instead of original price
- Added totalContributionUsd attribute in `sqlRowToItemObj` to a avoid calculation on front-end
- Split cause-specific bookkeeping into helper functions in `donationHelpers.js`

Remaining Todo's:
- Implement multiple-currency flexibility
- I couldn't figure out how to fix the dependency cycle error in `donationHelpers.js`
- Refactor out any uses of the `items` attribute of a donation object

lucashu1:
Summary of some additional changes since handoff:

* Show partially funded items in donation tracker
* Integrate Sendgrid monthly donation confirmation email
* Set minimum item price to $0.25
* Clean up some SQL pipe-delimiter code (use a `splitSqlCol` helper function)
* Added some `utils` in `packages/frontend` to clean up currency-conversion code (more to be done in a future PR)

Note: Left the `item.donor` object in for now (it uses the donor info from the fully-funded contribution), because it's still used in a lot of places, and would be a big pain to refactor out right now

packages/frontend/components/FamilyProfile/ItemCard.jsx
```diff
@@ -270,10 +275,7 @@ const ItemCard = ({
             <React.Fragment>
               <div className={classes.priceContainer}>
                 <div className={classes.price}>
-                  {`$${fx(need.price)
-                    .from('EUR')
-                    .to('USD')
-                    .toFixed(2)}`}
+                  {`$${remainingPriceUsd.toFixed(2)}`}
```
Raghav-Maheshwari, line 278: @lucashu1 can we also implement "crossing out" the original price as discussed here? https://giveduet.slack.com/archives/CJX92A10V/p1597634858005700?thread_ts=1597627086.001800&cid=CJX92A10V

I feel like this + the extra tooltip shouldn't be too much extra code and we can just bundle that update in with this PR instead of adding it later in a separate PR

packages/frontend/pages/success.jsx
```diff
@@ -29,8 +29,11 @@ class Success extends React.Component {
       return <Error statusCode={errorCode} />;
     }
 
-    // deep copy
-    const items = donation.items.map(item => ({ ...item }));
+    // get fully-funded items from donation.contributions
+    const items = donation.contributions
```
Raghav-Maheshwari, line 33: I don't think we currently route to the `/success` page from `/subscribe`, so not sure if filtering for fully-funded items is necessary here (as in, all the items in `donation.contributions` should be fully-funded because we're coming from the regular item flow)

lucashu1: You're right, it's not strictly necessary, but I added it just in case

packages/frontend/components/FamilyProfile/ItemCard.jsx
lucashu1: Yup! On it

lucashu1: Done @Raghav-Maheshwari! Check out the updated PR description.

lucashu1:
@Raghav-Maheshwari thoughts on merging this in now that the emails are ready to go? Imo, what we can do is merge it into master and double-check on staging that the webhook calls work correctly, and then once that looks good to go, redeploy to prod -- maybe next weekend? (Idk if I wanna deploy this to prod during my onboarding week lol, just in the off chance something goes wrong)

Raghav-Maheshwari (approved):

PR #62, opened 2020-08-03
Urgent items backend

lucashu1:
**Summary of changes:**
* New columns in `items` table and `items_view`: `urgent` (whether an item is currently marked as "urgent" or not), and `unsuccessful_urgent_attempt` (`true` if the beneficiary tried to request this item as `urgent` when initially filling out the Typeform, but already had the max number of urgent items)
* Add `urgent` to `item` data model
* Typeform webhook accepts an `urgent` question; update `urgent` and `unsuccessful_urgent_attempt` columns in database accordingly

(See "Urgent Items" section under beneficiary portal [product spec](https://docs.google.com/document/d/1xmIWvMJq1nW4Z7E-l_toJLCecF4d_H9Njla0IG8HgZo/edit#))

**Todo:**
* Add "Is this an urgent item?" question to Typeform in multiple languages (copy is in the Beneficiary Portal product spec)
* Create an API route to allow beneficiaries to update urgent items
* Testing!


packages/backend/src/server/util/typeformHelpers.js
```diff
@@ -6,11 +6,10 @@ async function insertItemFromTypeform(itemInfo) {
...
     const conn = await config.dbInitConnectPromise();
-    const [
-      results,
-    ] = await conn.query(
-      'INSERT INTO items (name,size,price_euros,beneficiary_id,category_id,comment,status,store_id,link,price_tag_photo_link,in_notification) ' +
-        'VALUES (?,?,?,?,?,?,?,?,?,?,?)',
+    const [results] = await conn.query(
+      `INSERT INTO items
+        (name,size,price_euros,beneficiary_id,category_id,comment,status,store_id,link,price_tag_photo_link,in_notification,urgent,unsuccessful_urgent_attempt)
+        VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?)`,
```
Raghav-Maheshwari, line 12: @lucashu1 is the `unsuccessful_urgent_attempt` column going to be used for anything?

lucashu1: We potentially could! If we have time for design + engineering, we could show an error notice when the user is redirected back to the beneficiary portal. But if we don't have time, we could just use it for our own internal analytics, to see how many beneficiaries are trying to request too many urgent items. Might hint that we have to reword our instructions to make them more clear

PR #63, opened 2020-08-05
Authorization HOCs & Utils

lucashu1:
**Summary:**
Use Higher Order Components (HOCs) to abstract away as much authorization logic as possible. Inspired by: https://github.com/vercel/next.js/discussions/10925#discussioncomment-12471

Creates 2 new auth-related wrappers, used similarly to `injectSheet(styles)(...)`:
* `withAuthServerSideProps(authOptions)(...)` for injecting `authData` as a second parameter into the lower-level `getServerSideProps(ctx, authData)` (server-side)
* `withAuth(authOptions)(...)` for handling redirects if not authorized (client-side)

**tl;dr** enforcing authorization (and handling redirects, etc.) is now as easy as adding a wrapper around `getServerSideProps`, and another wrapper around the Page-level `export`! And getting info on the current user just requires reading from the app-level `UserContext`.

**Backend:**
* New `/api/authenticateUser` API route: accepts a token in Authorization headers (using `bearer` scheme), and sends back an `authData` object: `{ loggedIn, user, userType, duetToken }` (replaces all `checkLogin` routes)
* `user` is either a `store` or `beneficiary` object, or a dummy `admin` object (which just contains the admin username)
* `userType` is one of `store`, `beneficiary`, or `admin`
* Remove all old `checkLogin` routes (no longer needed!)

**Frontend:**
* Helper `getAuthDataFromToken(duetToken)`: call backend, return `{ loggedIn, user, userType, duetToken }`
* Server-Side `withAuthServerSideProps`: inject `authData` object into lower-level `getServerSideProps`
* Client-Side `withAuthHOC`: takes in an `authOptions` object argument: `{ userType (string), redirectOnAuthFailure (string/path) }`, and handles redirects automatically
* `AuthContext` React context: makes `authData` visible to any lower-level components, and also allows any lower-level components to update the current login state via `updateAuthState(token)` and `logout()`. Handles cookie-setting, etc.
* Modify existing `checkLogin` authorization flows to use `withAuthServerSideProps` instead of making separate `checkLogin` calls
* Magic link support, via `withAuthServerSideProps`: e.g. if the user navigates to `giveduet.org/stores-portal/1?token={tokenHere}` `withAuthServerSideProps` will read the token, authenticate the user, and set a login cookie
* Stores portal and beneficiary portal login pages auto-redirect if already logged in

**Follow-on work:**
* Nice-to-have: Site-wide magic link support (so that we can authorize the user from any page URL where `?token` is included -- not just from pages with `withAuthServerSideProps` explicitly added). For now, I don't think this will be a problem, because the profile page (where the magic link will direct) will definitely use `withAuthServerSideProps`
* Add donor auth strategy (it probably makes more sense to have this be in a separate donor accounts-specific PR, so that we can fully test out magic links in one PR)

**Auth wrappers usage example:**
```
import React from 'react';

import Layout from '../../components/shared/layout';
import Portal from '../../components/StorePortal';

import { withAuth, withAuthServerSideProps } from '../../utils/auth';

const authOptions = {
  userType: 'store',
  redirectOnAuthFailure: '/stores-portal',
};

export const getServerSideProps = withAuthServerSideProps(authOptions)(
  async ({ query }, { user }) => {
    const { storeId } = query;

    const items = await fetch(
      `${process.env.DUET_API}/api/items?store_id=${storeId}`
    )
      .then(res => res.json())
      .then(itemsJSON =>
        itemsJSON.map(item => ({ ...item, isChecked: false }))
      );

    return {
      props: {
        storeId,
        items,
        user,
      },
    };
  }
);

class StorePortal extends React.PureComponent {
  render() {
    const { items, user } = this.props;

    console.log(`StorePortal user: ${JSON.stringify(user)}`);

    return (
      <Layout title="Duet | Store Portal">
        <Portal storeData={items} />
      </Layout>
    );
  }
}

export default withAuth(authOptions)(StorePortal);
```

**UserContext usage example:**
(Note that the `state` within `UserContext` might not yet be fully updated at initial render, so we have to check first for the existence of `user`. It seems to refresh pretty shortly thereafter, though.)
```
import UserContext from '../../../contexts/UserContext';

class BeneficiaryPortal extends React.PureComponent {
  render() {
    const { beneficiary } = this.props;
    const {
      state: { user },
    } = this.context;

    return (
      <Layout title="Duet | Beneficiary Portal">
        <div>{user && user.username}</div>
        <BeneficiaryPortalComponent beneficiary={beneficiary} />
      </Layout>
    );
  }
}

export default withAuth(authOptions)(BeneficiaryPortal);

BeneficiaryPortal.contextType = UserContext;
```


packages/backend/src/server/controller/auth.js
```diff
@@ -0,0 +1,36 @@
+async function authenticateUser(req, res) {
+  if (req.user) {
```
Raghav-Maheshwari, line 2: lets add a comment that `req.user` is coming from passport middleware

packages/frontend/contexts/UserContextProvider.jsx
```diff
@@ -0,0 +1,86 @@
...
+    const { user } = authData;
+    if (user) {
+      // if successfully logged in, set a cookie
+      setCookie(null, 'duetToken', token, { path: '/' });
+    }
+
+    console.log(
+      `UserContextProvider/updateAuthState: set current user to ${JSON.stringify(
+        user
+      )}`
```
Raghav-Maheshwari, line 47: let's not console log this because it will be visible on the client-side

PR #66, opened 2020-08-07
Use subscriptions table in database

lucashu1:
* Upon a new subscription, insert into `subscriptions` table
* Some minor donation-related code refactoring/cleaning


packages/backend/src/server/util/donationHelpers.js
```diff
@@ -223,7 +223,6 @@ async function insertDonationIntoDB({
     const { honoreeEmail, honoreeFirst, honoreeLast, honoreeMessage } =
```
Raghav-Maheshwari, line 223: can't directly comment on it, but line 218 still says: `donorInfo` as a function parameter attribute

PR #71, opened 2020-08-14
Manage Account Page

nicolengsy:
Basic React components and interactions completed, handing off to @lucashu1 and @Raghav-Maheshwari! ~~Since auth isn't working yet, subscription info on the page is all hard-coded.~~

See files in:
- `pages/account/[username]`
- `components/ManageAccountPage`

Summary of Todo's (I also added `TODO` comments in those files):
- ~~Add column to donors table to store bool of `showPublicImpactPage`~~
- ~~Get auth to work so GET and POST requests can be made, and implement these requests~~
- ~~After figuring out auth, dynamically populate subscription information~~
- ~~Check a user is logged in~~

### Changes since handoff -> Lucas:
* Working auth: for now, requires you to have a `token` in the query parameters (will be replaced with magic link auth)
* Fetch `donor` and `subscriptions` from backend, use to populate React `props`
* Redirect to landing page if user is not logged in
* Instead of cancelling subscriptions with the click of a button, tell them to email us instead
* Working `showPublicProfilePage` toggle
* Working "edit account info" (name, email) functionality, with form validation
* Mobile responsive

**Testing/Login Instructions:**
1. Log into Insomnia "Duet (Main)", and go to "Donors" folder -> "Donor login (WIP, Manage Acct Review App)"
2. Enter in your username in the request body, send the request
3. Get the token from the response
4. Go to https://duet-mono-website-pr-71.herokuapp.com/account/{usernameHere}?token={tokenHere}

### Screenshots:
**Main page:**
Screen Shot 2020-08-24 at 1 27 11 PM

**Email already taken by another donor:**
Screen Shot 2020-08-24 at 3 26 23 PM

**No subscriptions + "Private" setting:**
Screen Shot 2020-08-24 at 3 26 43 PM

**Subscription boxes:**
Screen Shot 2020-08-24 at 3 32 41 PM


packages/frontend/components/ManageAccountPage/index.jsx
```diff
@@ -0,0 +1,329 @@
...
+        this.setState({
+          error: 'Please provide a valid email address',
+        });
+        return false;
+      }
+      // if nothing has changed, then close form and do nothing
+      if (
+        firstName === donor.donorFirst &&
+        lastName === donor.donorLast &&
+        email === donor.donorEmail
```
Raghav-Maheshwari, line 137: found a bug here
1. Update the donor info i.e from Raghav > Ragh, 
2. Try to update it back from Ragh -> Raghav 
3. It won't call `api/modify` again because the `donor` object passed in as a prop doesn't change/update on the initial modify call, so we end up with "raghav" === "raghav" and we return/exit out of this function.

To fix this, you might want to keep track of the `donor` object coming in from the BE as a part of the `state` of this component, and when you call `api/modiy`, update the `state` to the donor obj that is returned from the `api/modiy` call.

packages/frontend/components/ManageAccountPage/index.jsx
```diff
@@ -0,0 +1,329 @@
...
+                  />
+                </div>
+              )}
+            </div>
+          </div>
+
+          {!editInfo && (
+            <React.Fragment>
+              <div className={classes.text}>
+                {formValues.firstName} {formValues.lastName}
```
Raghav-Maheshwari, line 253: imo I think we should render the donor info based on the `donor` prop (or state - see my comment above) instead of from `formValues`

packages/frontend/pages/account/[username]/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const getServerSideProps = withAuthServerSideProps(authOptions)(
+  async ({ query: { username } }, { duetToken }) => {
+    // get donor object
+    const donorObj = await fetch(
+      `${process.env.DUET_API}/api/donors/${username}`
+    ).then(res => res.json());
+
+    // get subscriptions for this donor
+    const subscriptionsRes = await fetch(
```
Comment (Raghav-Maheshwari) [inline packages/frontend/pages/account/[username]/index.jsx:22]: can we use `[donor, subscriptions] = Promise.all(...)` here? I think its a little faster than awaiting both fetches one after another

packages/frontend/components/ManageAccountPage/index.jsx
lucashu1: Good call, will update that

packages/frontend/components/ManageAccountPage/index.jsx
lucashu1: Good catch, thanks!

packages/backend/src/server/controller/donors.js
```diff
@@ -69,7 +69,8 @@ async function modifyDonor(req, res) {
       }
     }
     await donorHelpers.modifyDonorInfo(donorId, options); // req.user from passport middleware
-    return res.sendStatus(200);
+    const donor = await donorHelpers.getDonorObjFromDonorId(donorId); // get updated donor object
```
Raghav-Maheshwari, line 72: can we just have `donorHelpers.modifyDonorInfo` return the updated donor object? Then we don't have to do this extra call to `getDonorObjFromDonorId`

Raghav-Maheshwari (approved):

packages/backend/src/server/controller/donors.js
lucashu1: Yep, will do, and then merge this in

PR #89, opened 2020-09-18
Upgrade facebook messaging_referral API to v8.0

lucashu1:
In November, Facebook will be be pushing a break change to the messaging API: the chat-plugin will use a new `sender.user_ref` field instead of `sender.id` (`PSID`, or page-scoped user ID). This PR adds support for this new `user_ref` protocol.

See: https://developers.facebook.com/docs/messenger-platform/reference/webhook-events/messaging_referrals/

In the future, when we want to further automate refugee onboarding (perhaps through a signup form), we can use this messenger "checkbox" feature (which uses `user_ref`) in order to get the `user_ref` during the signup process: https://developers.facebook.com/docs/messenger-platform/discovery/checkbox-plugin/

**Summary of changes:**
* Use `request-promise` to send FB messages, instead of the outdated `fb-messenger` node package
* Add a `fb_user_ref` column to the beneficiaries table
* Add `fbUserRef` to the beneficiary data model
* On receiving a new message from the chat-plugin, update the `fb_user_ref` column in the `beneficiaries` table
* Support sending messages by either PSID or by user_ref (prioritizing PSID, if it exists)

Tested by both sending and receiving messages by `PSID` and by `user_ref`. Verified that messages get sent successfully, and that the `fb_user_ref` column gets updated for the corresponding beneficiary.

**Screenshot:**
Screen Shot 2020-09-17 at 9 15 44 PM


packages/backend/src/server/util/beneficiaryHelpers.js
```diff
@@ -17,6 +17,7 @@ function sqlRowToBeneficiaryObj(row) {
     story: row.story,
     language: row.language,
     fbPsid: row.fb_psid,
+    fbUserRef: row.fb_user_ref,
```
Raghav-Maheshwari, line 20: nit: this is outside the scope of this PR, but I feel like our beneficiary data model is becoming pretty big and unwieldy, especially if we're passing all of this data to the frontend and only a portion of it is needed at any given time.

It might be nice to try and refactor some of these data models out, or only send the FE the exact data that's needed? Most definitely something for the backlog or to ruminate on

Raghav-Maheshwari (approved):

packages/backend/src/server/util/beneficiaryHelpers.js
lucashu1: Definitely. Stuff like `fbPsid`, `fbUserRef` can just be internal to the backend, and not passed through the API itself. Maybe we could make more of the attributes nested (similar to how we handled `campaignInfo` in `captureObj`), and then only keep the high-level attributes that we actually need for any given use case

PR #90, opened 2020-09-20
Messages for meal and campaign based donations

dishamevada:
## **Description**
- Before, the donation tracker only showed messaged for item based donations
- Now it shows messages for meal and campaign based donations
## **How to Test**
- On your local machine go to 
   - http://localhost:3000/donation?donationId=399 for the meal based donation message 
   - http://localhost:3000/donation?donationId=318 for the campaign based donation message 
## **Screenshots** 
Screen Shot 2020-09-19 at 4 18 34 PM
Screen Shot 2020-09-19 at 4 18 50 PM


packages/frontend/pages/donation.jsx
```diff
@@ -14,11 +14,12 @@ class Home extends React.Component {
       return { errorCode: 404 };
     }
 
-    const [donation, currencyData] = await Promise.all([
+    const [donation, currencyData, campaign] = await Promise.all([
       fetch(`${process.env.DUET_API}/api/donations/${donationId}`).then(res =>
         res.json()
       ),
       fetch(`${process.env.DUET_API}/api/currency`).then(res => res.json()),
+      fetch(`${process.env.DUET_API}/api/campaigns`).then(res => res.json())
```
lucashu1, line 22: Instead of fetching all the campaigns, imo it might be better to just fetch the specific campaign associated with this donation. I think it's fine if that means that we have to have a separate `fetch` call after fetching the donation, instead of including the `campaigns` call in the `Promise.all(...)` call.

What you have here definitely works for now, but when we (one day!) have a whole lot of campaigns in our database, it'll probably be better to just fetch the single campaign.

If this donation _isn't_ associated with a campaign, then we can leave `campaign` as `undefined`.

packages/frontend/components/DonationTracker/donationTracker.jsx
```diff
@@ -119,6 +119,16 @@ class DonationTracker extends React.PureComponent {
         />
       ));
 
+    const numMeals = donation.donatedMeals.length;
```
lucashu1, line 122: Just in case `donatedMeals` is omitted or `undefined`, you might consider changing this to `donation.donatedMeals ? donation.donatedMeals.length : undefined`.

packages/frontend/components/DonationTracker/donationTracker.jsx
```diff
@@ -119,6 +119,16 @@ class DonationTracker extends React.PureComponent {
...
+    const isCampaignDonation = (donation.campaignId > 0) ? true : false;
+    let campaignName = "";
+    if(isCampaignDonation){
+      Object.values(campaign).forEach(campaignInfo => {
```
lucashu1, line 126: See other comment: if you only fetch the single campaign associated with this donation, then you won't need to do this iteration, since `campaign` will already represent just the single campaign object

packages/frontend/components/DonationTracker/donationTracker.jsx
```diff
@@ -119,6 +119,16 @@ class DonationTracker extends React.PureComponent {
         />
       ));
 
+    const numMeals = donation.donatedMeals.length;
+    const isCampaignDonation = (donation.campaignId > 0) ? true : false;
```
lucashu1, line 123: Ah! I think we've actually updated our data models/object definitions since the start of this PR. Instead of checking for `donation.campaignId`, you'll actually want to check for `donation.campaignInfo` now. I think that the `donation.campaignId` attribute has been deprecated. The API docs in Notion should be updated to reflect this.

Those changes are probably responsible for the merge conflicts too, which we should try to fix while we're at it. Happy to help you in resolving these conflicts if you have any questions.

lucashu1 (changes requested):
Thanks for getting this pushed @dishamevada!

I added some code-styling related comments to address. Hopefully those are helpful. Lmk if you have any questions about the merge conflicts, or if you need help resolving those!

packages/frontend/pages/donation.jsx
Raghav-Maheshwari: yep just echoing thoughts here - if we are going to name our prop `campaign`, then we anticipate that we are passing in a single campaign object, instead of an `array<campaign>`

packages/frontend/components/DonationTracker/donationTracker.jsx
```diff
@@ -108,7 +108,7 @@ const styles = {
 
 class DonationTracker extends React.PureComponent {
   render() {
-    const { classes, donation, family } = this.props;
+    const { classes, donation, family, campaign} = this.props;
```
Raghav-Maheshwari, line 111: can you add `campaign` as a prop type on line 13?
here's more info on react prop types: https://reactjs.org/docs/typechecking-with-proptypes.html

if you're feeling extra motivated, you can also add a proptype for `family` (we're currently missing one for it too)

Raghav-Maheshwari (changes requested):

lucashu1:
As discussed earlier today, marking this as a draft for now until we make a slight revision to the `campaigns` API: https://public.3.basecamp.com/p/c9rG8wYNK7y83fKri58rT9W5

PR #91, opened 2020-09-20
Refugee Donor Impact Page

Raghav-Maheshwari:
This PR:
- Implements the public donor impact page for the refugee cause
- available at `/profile/{donorUsername}`

Screenshots:
**Desktop**
image

**Mobile:**
image

Outstanding (What this PR doesn't do):
- Block requests to the page if the impact page is hidden (based on donor account settings)
- Implement "your donations" card
- Implement refer a friend functionality (designs pending)
- Implement campaign or meals funded cards (designs pending)
- Dynamic meta tags for donor accounts

@lucashu1 after developing this page, I've come to realize that a lot of UI components are missing for donors who never donated to the refugee cause... I feel like we have two options: only push the donor impact page to refugee donors, or wait to fully roll it out until designs for meals and other campaigns are complete (cc @spenceblood )


lucashu1:
@Raghav-Maheshwari Good point: regarding "only pushing the donor impact page to refugee donors", do you have any thoughts on how we'd do that? Would it work if we set the `show_public_profile_page` flag to `false` for non-refugee donors?

In those situations, maybe we could also show a simple placeholder message like: "You haven't donated any items to our refugee families yet! Click [here](https://giveduet.org/give) to support one of our partner families".

Also definitely wanna hear @spenceblood's thoughts

lucashu1:
Also, I just wanna say that this page looks awesome and it's gonna be super cool when it's launched

PR #92, opened 2020-09-21
Add 404 not found page

zero-delta:
### Description

- Add 404 not found page 
- Available at `/pages/404`

### Screenshots
notfoundpage


zero-delta:
Replacing NextJs Error component with custom 404 NotFound component in `[...slug].jsx` 
Now uses the custom NotFound page for all 404 status codes.

lucashu1:
@zero-delta The main 404 functionality is working for me! However, the error message looks pretty weird when rendered on Safari (see screenshot). I think it might be due to the usage of `<h1>`. Instead of `<h1>`, maybe just try using a regular `div` with CSS styling for the text instead?

Nit: if the rest of the text is bold, we might wanna make the "here" link bold as well.

**Safari:**
Screen Shot 2020-09-21 at 7 54 57 PM

**Inspect Element:**
Screen Shot 2020-09-21 at 7 57 56 PM

lucashu1 (changes requested):
@zero-delta The main 404 functionality is working for me! However, the error message looks pretty weird when rendered on Safari (see screenshot). I think it might be due to the usage of `<h1>`. Instead of `<h1>`, maybe just try using a regular `div` with CSS styling for the text instead?

Nit: if the rest of the text is bold, we might wanna make the "here" link bold as well.

**Safari:**
Screen Shot 2020-09-21 at 7 54 57 PM

**Inspect Element:**
Screen Shot 2020-09-21 at 7 57 56 PM

packages/frontend/pages/404.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+const propTypes = {
+  classes: PropTypes.object.isRequired,
+};
+
+const styles = {
+  container: {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
```
Raghav-Maheshwari, line 20: nit: not sure if you actually need to use flexbox here - you should be able to achieve what you want with `textAlign: 'center'`

packages/frontend/pages/404.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+class NotFound extends React.PureComponent {
+  render() {
+    const { classes } = this.props;
+    return (
+      <Layout title="Duet | Not Found">
+        <NavBar />
+        <div className={classes.container}>
+          <h1>
+            {' '}
```
Raghav-Maheshwari, line 44: the linter might have added the `{' '}`, but could we try removing them? If not, you can use `&nbsp;` as a space

Raghav-Maheshwari (approved):
left a couple of nits - once you address Lucas's comments, it looks good to me!

packages/frontend/pages/404.jsx
zero-delta: When I try `textAlign: 'center'` : 
image
it ends up moving it to the top:
noflex

I tried adding paddingTop and some other stuff, but the text ends up going to the next line.
image
image

zero-delta:
Replaced `<h1>` with `<div>` and some font formatting
Safari:
safari7
Chrome:
chrome

lucashu1:
@zero-delta I was still getting the weird separated-text error when I shrunk my screen size so that the text didn't fit onto a single line.

Screen Shot 2020-09-23 at 4 49 09 PM

In order to fix this, I made some tweaks to the HTML/CSS. The main changes to fix that bug were:
* Put all text in a single `div` so that `flex-wrap` doesn't separate `divs` when text doesn't fit onto a single line
* Remove `flexWrap`
* Add `textAlign: center`
* Added `flexDirection: column` (although idk if this really did anything)

Additionally: 
* Added `marginTop: 80` (to leave room for absolutely positioned NavBar)
* Add padding to the left and right sides of the container (20px)

Be sure to take a look at the updated code and make sure you know what each of these properties are doing! Wanted to free up your queue before we create more assignments for this upcoming sprint.

In order to catch visual bugs like this, we'll typically use "Inspect Element" to inspect the page on a variety of screen sizes, since a lot of our donors visit the site on mobile as well. So we need to make sure that things work on a variety of screen sizes.

lucashu1 (approved):

